PR #31, opened 2024-09-12
Added GoalAnalysisConnector.

bre1470:
- [ ] Solve issue with authorization 102 (maybe wrong endpoint path?)


src/GoalAnalysis/GoalAnalysisOptions.ts
```diff
@@ -0,0 +1,149 @@
...
+     * Returns time series data for charting purposes.
+     */
+    includeDetailedInvestmentGrowthGraph?: boolean;
+
+    requestProbability?: number;
+
+    /**
+     * Investor’s target goal amount.
+     */
+    target?: number;
```
Eskils, line 132: Morningstar’s API docs say:

> **Note** Required when `IncludeDetailedInvestmentGrowthGraph` is set to `true`.

This is the case for parameters: `target`, `currentSavings`, `annualInvestment` and `timeHorizon`.

I would expect our documentation to say the same thing.

Eskils (approved):
This looks good. 🙌

I had a small nitpick on documentation.

PR #32, opened 2024-09-12
Added versioning.

bre1470:


src/version.ts
```diff
@@ -0,0 +1,2 @@
+export const version = '0.4.1';
```
Eskils, line 1: I would add a comment to this file saying that it was generated from a script.

Alternatively, say this in the file path. For instance: _src/generated/version.ts_

Eskils (approved):
This is nice to have! ⭐️

PR #34, opened 2024-09-13
Added CLI command for API docs.

bre1470:
Use `npx connectors-morningstar api`.


Eskils (changes requested):
This is a big PR, so I spent more time testing rather that looking though code.

This is great because it looks and works like the rest of the Highcharts API Reference, and also works offline.

## Requested changes:

Some things on the API docs website at localhost:8005 did not work like I expected:

- Navigating directly to an option (e.g. http://localhost:8005/RNANews.localization.currency) does not work. The home page is shown instead. This also means the links in the div#body does not work.

- The class reference does not exist. 
   + If this is not to be implemented: Consider hiding this link.
   + If it is to be kept: consider a different text. It now reads `For modifying the chart at runtime. See the [class reference](http://localhost:8005/class-reference/classes.list).`
Screenshot 2024-09-20 at 07 49 25

- Missing copyright year and version in footer
Screenshot 2024-09-20 at 07 50 56

- The search field does not work. Console shows `404 search.json`
Screenshot 2024-09-20 at 13 26 23

- Clicking “JS API Reference” in the top header links to the incorrect page which does not exist:
Screenshot 2024-09-20 at 13 43 13

- Edit button on the options link to the wrong page on GitHub: https://github.com/highcharts/highcharts/blob/vundefined/src/RNANews/RNANewsOptions.ts

- GitHub logo/link in the footer links to Highcharts repository. I expected it to link to the `connectors-morningstar` repository.

Eskils (approved):
This is excellent – well done!

PR #38, opened 2024-09-18
Added RiskScoreConnector

Eskils:
This feature branch implements a connector for Risk Score.

It’s more important that this feature is merged, then a more fancy demo can be implemented later. 
- Will look into making a better demo in my next PR.


src/RiskScore/RiskScoreConverter.ts
```diff
@@ -0,0 +1,230 @@
...
+            for (let columnIndex = 0; columnIndex < riskScoreValues.length; columnIndex++) {
+                const column = `${portfolio.name}_${valueColumns[columnIndex]}`;
+                const value = riskScoreValues[columnIndex];
+                table.setCell(column, Number(rowIndex), value);
+            }
+        }
+
+        if (errors.length > 0) {
+            throw new Error(errors.join('\n'));
+        }
```
Eskils, line 217: Should this be handled differently?

This is the errors from invalid holdings. Morningstar still gives a result, and throwing the errors here will make the user unable to show the result.

Using console.error or equivalent makes the developer unable to handle the error in their own way.

Is there already a different way to signal errors that will not stop execution? If not, is it worth implementing?

bre1470: If the errors are part of the JSON, you can add them to a override of RiskScoreConnector.metadata. So you have to add an errors property to the converter to pass it on. Also you should stick to the original JSON error message.

bre1470 (approved):
Great work! 👍

PR #47, opened 2024-11-14
Corrected URLs.

bm64:
Corrected URLs in docs and demos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,6 @@ The connectors are available as
 
 <!-- Link References -->
 
-[Highcharts Connectors]: https://highcharts.com/connectors/
+[Highcharts Connectors]: https://www.highcharts.com/products/morningstar-data/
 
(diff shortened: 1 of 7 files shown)


pawelfus (approved):
Thanks! 

Note: At this moment, there's no `/connectors` landing page, so we will redirect to Morningstar subpage. It will change once the landing page be ready. 

Edit: Anything to add @bre1470?

bre1470 (approved):
Looks good! 👍 

We could create a landing page for `code.highcharts.com/connectors/` as an alternative.

pawelfus (changes requested):
Found one more: https://github.com/highcharts/connectors-morningstar/blob/corrected-docs-demos-urls/docs/connectors/morningstar/time-series/time-series.md -> Regulatory News Announcements is broken

PR #63, opened 2024-12-27
feature/demo-page

bm64:
Implement designs for morningstar connector demos page.


pawellysy (approved):

kamil-musialowski (approved):

pawelfus (changes requested):
Nice! A few things, see my comments inline.

In addition: Could you add fetching the env file just once, then store it in the localStorage? So we don't have to search for the env file when scanning through demos. I mean: whichever demo loads the env file first, store it in the localStorage. And each demo should first check localStorage for this JSON file. If exists, load it, create chart and hide the button and description about uploading the file.

DJTechnoo (approved):
Looks very gorgeous!

pawelfus (approved):
Thanks!

PR #69, opened 2025-02-03
Added SecurityCompareConnector

DJTechnoo:
Added `SecurityCompareConnector`.


jedrzejruta (changes requested):
Overall looks really solid, I like the concept of defaultly pulling TrailingPerfomance :tada:

Q: When pulling assetAllocations and other types, what about the not-classified number? In XRay, we put it into the column name which isn't ideal, but we need to have this informations somewhere, perhaps in metada?

demos/security-compare-asset-allocations/demo.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <link rel="stylesheet" href="demo.css" />
+        <script src="https://code.highcharts.com/dashboards/datagrid.js"></script>
+        <script src="https://code.highcharts.com/highcharts.js"></script>
```
pawelfus, line 7: Please add a11y module and remove redundant scripts (eg grid). The same about CSS stylesheets

demos/stock-security-compare/demo.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <script src="https://code.highcharts.com/stock/highstock.js"></script>
```
pawelfus, line 5: Please add a11y modules

demos/security-compare-asset-allocations/demo.html
pawelfus: Edit: Or add Grids :)

demos/security-compare-asset-allocations/demo.css
```diff
@@ -0,0 +1,2 @@
+@import url("https://code.highcharts.com/css/highcharts.css");
```
jedrzejruta, line 1: One last thing - newest version of Dashboards doesn't enable `styledMode` in charts by default, so either remove the CSS import or enable `styledMode` in the chart components.
image

jedrzejruta (changes requested):
Nice work, only a minor thing in the demo, other than that looks great!

jedrzejruta (approved):

demos/security-compare-asset-allocations/demo.js
```diff
@@ -0,0 +1,122 @@
...
+        '99': 'Unclassified'
+    };
+
+    const result = ids.map(id => 
+        connector.table.getRowObjects()
+            .map(item => ({
+                name: typeMapping[item['AssetAllocations_Type_' + id]],
+                y: item['AssetAllocations_MorningstarEUR3_N_' + id]
+            }))
+            .filter(entry => entry.y !== 0) // Filter out entries where y is 0
```
pawelfus, line 40: I think we have to keep all values, otherwise we have mismatching colors in series (compare cash): 

Zrzut ekranu 2025-02-18 o 13 07 16

demos/security-compare-asset-allocations/demo.js
```diff
@@ -0,0 +1,122 @@
...
+async function displaySecurityDetails (postmanJSON) {
+    const ids = ['F000015CH2', 'F0000143IP'],
+        idNames = {
+            'F000015CH2': 
+                'Schroder International Selection Fund Global Emerging Market Opportunities',
+            'F0000143IP': 'Fidelity Funds - Asia Pacific Opportunities Fund I-Acc-USD'
+        }
+
+
+    const connector = new HighchartsConnectors.Morningstar.SecurityCompareConnector({
```
pawelfus, line 11: Let's move the connector config under Dash ataPool's config

demos/security-compare-asset-allocations/demo.js
```diff
@@ -0,0 +1,122 @@
...
+        '4': 'Other Instruments',
+        '99': 'Unclassified'
+    };
+
+    const result = ids.map(id => 
+        connector.table.getRowObjects()
+            .map(item => ({
+                name: typeMapping[item['AssetAllocations_Type_' + id]],
+                y: item['AssetAllocations_MorningstarEUR3_N_' + id]
+            }))
```
pawelfus, line 39: I think we can resign from this. Use `dataLabels.format` to apply the translation and columnAssignment to choose the right columns for charts

demos/security-compare-asset-allocations/demo.js
pawelfus, line 1: Directory name:
Let's follow the convention, and since we use Dash here, the directory name should be `dashboards-...`

src/SecurityCompare/Converters/TrailingPerformanceConverter.ts
pawelfus, line 1: I still believe this might be optimized. SecurityDetails parses one security. This one parses multiple, just using a bit different code. However under the hood it's the same:
```
const securityDetails = json[0]; // Details
...
```
vs
```
for (let i = 0; i < json.length; i++) {
    const securityCompare = json[i], // Compare
    ...
```

The way I see it (please correct me if I'm wrong!):
- use only exiting TrailingConverter
- if JSON contains more than one security, then column names are `<name>_MSID`, if just one, then just `<name>`
-  `metadata.ids` and `metadata.isins` are set for multi-security response. For one security, set `metadata.id` and `metadata.isin`

pawelfus: Note: I didn't check other converters, but perhaps the same can be solution can be used?

pawelfus (changes requested):
Thanks! Just a few demo-polishing things and one parser-related discussion

demos/dashboards-compare-asset-allocations/demo.js
```diff
@@ -0,0 +1,166 @@
...
+
+    const ids = ['F000015CH2', 'F0000143IP'],
+        idNames = {
+            'F000015CH2': 
+                'Schroder International Selection Fund Global Emerging Market Opportunities',
+            'F0000143IP': 'Fidelity Funds - Asia Pacific Opportunities Fund I-Acc-USD'
+        };
+
+    const commonTooltip = `
+    <span">
```
bm64, line 23: Typo?

bm64 (approved):
Small typo, other than that looks good.

src/SecurityDetails/Converters/AssetAllocationsConverter.ts
```diff
@@ -96,48 +100,56 @@ export class AssetAllocationsConverter extends SecurityDetailsConverter {
...
+        if (json.length > 1){
+            isCompare = true;
+        }
+        // Create tables
+        for (const security of json) {
+            const id = security.Id,
+                isin = security.Isin,
+                assetAllocations = security.Portfolios[0].AssetAllocations,
+                assetAllocationsTypeStr =
+                'AssetAllocations_Type' + (isCompare? `_${id}` : '');
```
jedrzejruta, line 116: Really nitpicking here, but I believe that there should be a space after variable when using ternary operator.
```suggestion
                'AssetAllocations_Type' + (isCompare ? `_${id}` : '');
```

src/SecurityDetails/Converters/AssetAllocationsConverter.ts
```diff
@@ -96,48 +100,56 @@ export class AssetAllocationsConverter extends SecurityDetailsConverter {
...
+            ids.push(id);
+            isins.push(isin);
 
             for (let i = 0; i < assetAllocations.length; i++) {
-                const asset = assetAllocations[i];
-
-                table.setColumn(`AssetAllocations_${asset.Type}_${asset.SalePosition}`);
+                const asset = assetAllocations[i],
+                    assetAllocationsAssetStr =
+                    `AssetAllocations_${asset.Type}_${asset.SalePosition}` + (isCompare? `_${id}` : '');
```
jedrzejruta, line 126:
```suggestion
                    `AssetAllocations_${asset.Type}_${asset.SalePosition}` + (isCompare ? `_${id}` : '');
```

jedrzejruta (changes requested):
Tremendous work, thanks!

Minor change with the space in ternary operator, other than that looks good to me!

src/SecurityCompare/SecurityCompareConnector.ts
```diff
@@ -0,0 +1,176 @@
...
+        await super.load();
+
+        const userOptions = { ...this.options, ...options };
+        const { security, viewIds = 'MFsnapshot' } = userOptions;
+        const { ids: securityIds, idType: securityIdType } = (security || {});
+        const api = this.api = this.api || new MorningstarAPI(userOptions.api);
+        const url = new MorningstarURL('ecint/v1/multi-securities/', api.baseURL);
+
+        const searchParams = url.searchParams;
+        searchParams.set('ids', '' + securityIds.join('|'));
```
pawelfus, line 135: @bm64 should we extract UTF_PIPE variable to a common place? Currently stored in Screener, perhaps better make it a utils const to reuse.

tests/SecurityDetails/SecurityDetailsConnector.test.ts
```diff
@@ -28,8 +28,8 @@ export async function securityDetailsLoad (
     Assert.deepStrictEqual(
         connector.table.getColumnNames(),
         [
-            'SecurityDetails_TrailingPerformance_TimePeriod',
-            'SecurityDetails_TrailingPerformance_Value'
+            'TrailingPerformance_TimePeriod',
+            'TrailingPerformance_Value'
```
pawelfus, line 32: We should also update docs. That it is `TrailingPerformance_TimePeriod`, not `SecurityDetails_TrailingPerformance_TimePeriod`

src/SecurityDetails/SecurityDetailsOptions.ts
```diff
@@ -64,13 +68,31 @@ export interface SecurityDetailsMetadata extends MorningstarMetadata {
...
+
 
 export interface SecurityDetailsOptions extends MorningstarOptions {
     security?: MorningstarSecurityOptions,
     viewId?: string,
     converter?: SecurityDetailsConverterOptions
 }
 
+
+interface SecurityCompareSecurity extends Omit<MorningstarSecurityOptions, 'id'> {
```
pawelfus, line 84: Q (everyone):
Not convinced here, depends what the team thinks of it:
Should we follow the hierarchical extending approach, instead of Omits? We have full control over interfaces here. At the same time it is only about removing `id` and adding `ids` 🤔

pawelfus (changes requested):
Amazing work! Only a few minor things to discuss (and possible improve).

Note:
It seems to me we can refactor even further, perhaps reducing number of converters to one, more generic one, where all we have to do is define path and column names (eg `columnNames = ['CountryExposure_Assets', 'CountryExposure_NotClassified', ... ]`) and it's counterparts in JSON. It won't be part of this PR, just something that might pop up in v2.0, together with XRay Converters' refactor and merge ([link](https://app.asana.com/1/222651295669536/project/1209509376623362/task/1209423584099867))).

src/SecurityDetails/SecurityDetailsOptions.ts
pawelfus: @kamil-musialowski @bm64 @jedrzejruta

src/SecurityDetails/Converters/CountryExposureConverter.ts
```diff
@@ -96,48 +100,72 @@ export class CountryExposureConverter extends SecurityDetailsConverter {
...
 
-        if (json.length) {
+        if (!json.length) {
+            return;
+        }
 
-            // Update table
+        if (json.length > 1) {
+            isCompare = true;
+        }
```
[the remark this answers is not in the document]
kamil-musialowski: To keep up with the convention in other converters, I would say it could be extracted 👍

src/SecurityDetails/SecurityDetailsOptions.ts
kamil-musialowski: I don't mind using omit in MS repo, I think it's quite clean.

kamil-musialowski (approved):
I've tried to nitpick something but couldn't - very good job! 

Just a small comment about handling error, the rest of the stuff has been picked up by colleagues.

src/SecurityCompare/SecurityCompareConnector.ts
```diff
@@ -0,0 +1,176 @@
...
+
+
+    /* *
+     *
+     *  Constructor
+     *
+     * */
+
+
+    public constructor (
```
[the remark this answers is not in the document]
DJTechnoo: _Note: TBC, possibly refactor with an AbstractSecurityDetails_

src/SecurityCompare/SecurityCompareConnector.ts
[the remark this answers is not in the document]
DJTechnoo: _Note: TBC, possibly refactor with an AbstractSecurityDetails_

src/SecurityDetails/Converters/CountryExposureConverter.ts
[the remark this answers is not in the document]
jedrzejruta: I agree with Kamil, if there's possibility that we might use it somewhere else (and also for keeping consistency) - let's go with proposed solution.

jedrzejruta (approved):
Great work, thanks! 🎉

pawelfus (approved):
Thanks!

src/SecurityCompare/SecurityCompareConnector.ts
bm64: Yes, I think we should do that, and same goes for `UTF_COLON`.

pawelfus: So last request from us @DJTechnoo :D

bre1470 (commented):
~Two tests~ This fails for me: ~RNANewsConnector.rnaNewsLoad and~ SecurityCompareConnector.countryExposureLoad

bre1470 (changes requested):
Great work. The approach with the converters looks good, but there is a need to either mix better or separate better.

* If it is just a variant of `SecurityDetails` then it would be enough to just extend the `SecurityDetailsConnector` instead of `MorningstarConnector`.

* If it is its own thing, then `SecurityCompareJSON.ts` and `SecurityCompareOptions.ts` are missing and the connector separation has to be more consistent.

Therefore not every of my comments is valid, as it depends on which path you like to follow.

Remember to make use of converter options and maybe move some compare-related code to the connector as the connector has access to JSON and metadata as well.

src/SecurityDetails/Converters/AssetAllocationsConverter.ts
```diff
@@ -96,48 +100,57 @@ export class AssetAllocationsConverter extends SecurityDetailsConverter {
...
-            metadata.isin = securityDetails.Isin;
         }
 
+        if (isCompare){
+            (metadata as SecurityCompareMetadata).ids = ids;
+            (metadata as SecurityCompareMetadata).isins = isins;
+        } else {
+            (metadata as SecurityDetailsMetadata).id = ids[0];
+            (metadata as SecurityDetailsMetadata).isin = isins[0];
+        }
```
bre1470, line 152: ... the rest was not saved. This can become a mixed Metadata interface and you simply set all four props in the same object plus the `isMultipleDetails` flag.

src/Shared/Utilities.ts
```diff
@@ -0,0 +1,2 @@
+export const UTF_PIPE = '|';
+export const UTF_COLON = ':';
```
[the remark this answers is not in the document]
pawelfus: It started with some bug in Mstar API, that we had to escape this char, then we had to switch back to the current one. Instead of changing this in all files (and risking missing any) we decided to go with one-to-rule-them-all variable.

docs/connectors/morningstar/security-compare.md
```diff
@@ -0,0 +1,92 @@
+# Security Compare
+
+This type retrieves investment data for multiple securities, making it ideal for comparison.
+
+## How to use SecurityCompare
+
+Use the `SecurityCompareConnector` to load the connector.
+
+Specify the securities in the options along with a Postman environment file for authentication and other parameters such as `currency`.
```
bre1470, line 9: A postman file is not a must. In implementations you can also use a session token, or just form fields to pass username and password on. This should be clarified.

bre1470 (approved):
One thing regarding docs to consider, otherwise it looks much better. 👍

TorsteinHonsi (approved):
Thanks!

PR #86, opened 2025-04-25
Added support for loading env file from localstorage.

### LINKED ISSUE #72: Demo pages improvements
To do:

- [ ] - once the env file is being loaded, keep it in memory, then reuse in other demos (right now, each time you change demo, you need to provide this env file)

kamil-musialowski:
Fixes #72.

Internal note: For now it's just implemented in one demo. Once accepted, I'll just copy it over to other demos.

### DIFF
--- a/demos/stock-timeseries/demo.js
+++ b/demos/stock-timeseries/demo.js
@@ -1,65 +1,90 @@
-document.getElementById('postman-json').addEventListener(
-    'change',
-    async function (evt) {
-        const target = evt.target;
-        const postmanJSON = await getPostmanJSON(target);
+window.addEventListener('DOMContentLoaded', async () => {
+    const parsedPostman = JSON.parse(localStorage.getItem('postmanEnvironment')),
+        postmanMessage = document.getElementById('postman-message');
 
-        if (!postmanJSON) {
-            return;
-        }
+    try {
+        if (HighchartsConnectors.Morningstar.Shared.isPostmanEnvironmentJSON(parsedPostman)) {
+            initializeChart(parsedPostman);
+            postmanMessage.style.display = 'none';
+        } else {
+            localStorage.removeItem('postmanEnvironment');
 
-        target.parentNode.style.display = 'none';
+            const fileInput = document.getElementById('postman-json');
+            postmanMessage.style.display = 'block';
 
-        const priceConnector = new HighchartsConnectors.Morningstar.TimeSeriesConnector({
-            postman: {
-                environmentJSON: postmanJSON
-            },
-            series: {
-                type: 'Price'
-            },
-            securities: [{
-                id: 'US0378331005',
-                idType: 'ISIN'
-            }],
-            startDate: '2020-01-01',
-            endDate: '2020-12-31',
-            currencyId: 'EUR'
-        });
+            fileInput.addEventListener('change', async function (evt) {
+                const target = evt.target;
+                const postmanJSON = await getPostmanJSON(target);
 
-        await priceConnector.load();
-
-        Highcharts.stockChart('container', {
-            title: {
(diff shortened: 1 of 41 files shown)


demos/stock-timeseries/demo.js
```diff
@@ -1,65 +1,90 @@
...
+
+                localStorage.setItem('postmanEnvironment', JSON.stringify(postmanJSON));
+
+                initializeChart(postmanJSON);
+                postmanMessage.style.display = 'none';
+            });
+        }
+    } catch (error) {
+        // eslint-disable-next-line no-console
+        console.log('Incorrect postman environment file');
```
pawelfus, line 31: use console.error + include the error in the message

pawelfus (changes requested):
Nice! One minor thing, see below.

I have one question though, to discuss together with the team: would it make sense to move that utility part (loading env + localstorage usage) to an external script? It would simplify the demo. As a result we could have something like this:

```ts
/*
Load from localstorage, or
Setup change listener and resolve the promise once file is loaded
*/
const env = await getMorningstarEnvFile();

initializeChart(env);
```

kamil-musialowski:
Yes, I thought about moving it to an external script, but to be honest, I have no strong opinion on that. 

The code I've added also relies on the ID of HTML element to toggle it's visibility, so if we would like to have it in an external script, the ID of the "add file" message would have to always be the same.

But I'm open for the discussion and suggestions!

jedrzejruta:
I'm into the idea of extracting it into separate script, I feel like the demos could become much smaller in size of config and that's a huge benefit.

> The code I've added also relies on the ID of HTML element to toggle it's visibility, so if we would like to have it in an external script, the ID of the "add file" message would have to always be the same.

Technically, it shouldn't be a problem since we're switching between demos in iframe, so only one iframe is visible and loaded at a time (meaning no duplicate ID's shown at the same page). But for the clarity, we could also use a shared CSS class instead.

demos/stock-timeseries/demo.js
```diff
@@ -1,65 +1,90 @@
-document.getElementById('postman-json').addEventListener(
-    'change',
-    async function (evt) {
-        const target = evt.target;
-        const postmanJSON = await getPostmanJSON(target);
+window.addEventListener('DOMContentLoaded', async () => {
```
jedrzejruta, line 1: Maybe I'm missing something but why not use the `document` instead of `window`?

jedrzejruta (changes requested):
Nice work, only one question, other than that looks great! :clap:

DJTechnoo (approved):
Thanks, looking forward to finally have this!

bm64 (approved):
I also think that the utility part should be extracted to external script. Other than that looks good 👍

kamil-musialowski:
It definitely looks cleaner and better now, thank you for the suggestion. Here are the open questions:

- is the current structure, naming scheme, and new file naming and location fine?
- what about the `'postman-message'` and `'postman-json'` IDs, should I leave it as it is, or add new params to the function and allows users to pass the IDs through `getPostmanFile(initializeChart, postmanMessageID, postmanJSONID)`?

DJTechnoo (approved):

jedrzejruta (approved):
Nicely done! :clap:

Personally I thing that the naming convention is fine and should be the same across all demos.

pawelfus (approved):
Looks good, I think we can go with that solution.

kamil-musialowski:
Thank you, everything is implemented now and works beautifully. I've left the manual fetch demo untouched, to keep it manual 😄

pawelfus (approved):
Thanks!

bm64 (approved):

PR #87, opened 2025-04-25
Added dataModifier support to Morningstar connectors.

jedrzejruta:
Added dataModifier support to Morningstar connectors.

### DIFF
--- a/src/Shared/External.ts
+++ b/src/Shared/External.ts
@@ -57,10 +57,10 @@ export type DataTable = _DataTable;
  * */
 
 
-const Dashboards: typeof _Dashboards = globalThis.Dashboards;
+const Dashboards: typeof _Dashboards = globalThis.window.Dashboards;
 
 
-const DataGrid: typeof _DataGrid = globalThis.DataGrid;
+const DataGrid: typeof _DataGrid = globalThis.window.DataGrid;
 
 
 export const DataConnector = (
--- a/tests/GoalAnalysis/GoalAnalysisConnector.test.ts
+++ b/tests/GoalAnalysis/GoalAnalysisConnector.test.ts
@@ -1,4 +1,5 @@
 import * as Assert from 'node:assert/strict';
+import '@highcharts/dashboards/es-modules/masters/dashboards.src';
 import * as MC from '../../code/connectors-morningstar.src';
 
 export async function goalAnalysisConnector (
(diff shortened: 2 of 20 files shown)


tests/GoalAnalysis/GoalAnalysisConnector.test.ts
```diff
@@ -1,4 +1,5 @@
 import * as Assert from 'node:assert/strict';
+import '@highcharts/dashboards/es-modules/masters/dashboards.src';
```
pawelfus, line 2: Q: Why this one test needs to import this, while others don't?

src/Shared/External.ts
```diff
@@ -57,10 +57,10 @@ export type DataTable = _DataTable;
  * */
 
 
-const Dashboards: typeof _Dashboards = globalThis.Dashboards;
+const Dashboards: typeof _Dashboards = globalThis.window.Dashboards;
 
 
-const DataGrid: typeof _DataGrid = globalThis.DataGrid;
+const DataGrid: typeof _DataGrid = globalThis.window.DataGrid;
```
pawelfus, line 63: Q: Could you explain this change to me?

pawelfus (approved):
Looks really good. Did you consider a one generic test (eg in `shared`) to iterate over each connector? That would test future connectors without extra effort. Something like this:

```
for (var connector of MC) {
  if (connector instance of MC.MorningstarConnector) {
    ...
  }
}
```

Just a discussion point (thus accepting the PR). As of now, there is only two new endpoints planned that we are aware of (Equity and ESG).

tests/GoalAnalysis/GoalAnalysisConnector.test.ts
jedrzejruta: I put it there because this test is the first one to get executed. After including it there, Dashboards reference is stored in the `window.Dashboards` object globally, meaning that it doesn't require importing further, unlike with the connectors itself.

src/Shared/External.ts
jedrzejruta: Basically, to make dataModifiers work with the connectors, it's necessary to either import Dashboards or DataGrid. In the browser environment, this approach worked correctly, but somehow in the Node.js, `globalThis` didn't include these products, resulting in `DataModifier` and it's types being undefined. Referencing them from `globalThis.window` solves the issue.

jedrzejruta:
> Did you consider a one generic test (eg in shared) to iterate over each connector? That would test future connectors without extra effort.
It would be great to have one, but the config for almost every connector differs so much that for now I've decided to simply add another asserts into the connectors tests.

tests/GoalAnalysis/GoalAnalysisConnector.test.ts
pawelfus: It is as long as we don't implement ESG connector, right? Or any other with lower value of the first ASCII char? Perhaps it should be imported in `prepareGlobals`, like other globals??

src/Shared/External.ts
pawelfus: Okay, thnaks!

tests/GoalAnalysis/GoalAnalysisConnector.test.ts
jedrzejruta: You're right, once the order changes it might result in failing tests. I think the best way would be to import it after the `prepareGlobals` method is called, so that the Dashboards can be initialized correctly.

bre1470 (approved):
LGTM 👍

DJTechnoo (approved):
Absolutely well done!

kamil-musialowski (approved):
Nice one! Congratulations 🥳

KamilKubik:
LGTM, nice work!

PR #89, opened 2025-05-09
Added breakdown converters

DJTechnoo:
Added the following new converters for `SecurityDetails` and `SecurityCompare`: `IndustryBreakdown`, `IndustryGroupBreakdown`, and `BondStatistics`.


pawelfus (commented):
Amazing work! A couple of qu4stions:

1. What do you think about another abstract layer in the converters? Currently Meta and Industry converters looks 99% the same?
2. It is possible to refactor CountryBreakdown too?  To match `getBreakdow()` approach.

DJTechnoo:
> Amazing work! A couple of qu4stions:
> 
> 1. What do you think about another abstract layer in the converters? Currently Meta and Industry converters looks 99% the same?
> 2. It is possible to refactor CountryBreakdown too?  To match `getBreakdow()` approach.

1. Definitely worth investigating, I believe it can be possible to get an abstract place for them.
2. `AssetAllocations` and `RegionalExposure`can be refactored. However `GlobalStockSectorBreakdown` has an additional property that would require more arguments and tweaks in the `getBreakdown` function.

src/Shared/SharedSecurityDetails.ts
```diff
@@ -26,13 +26,18 @@ import {
     CountryExposureConverter,
     PortfolioHoldingsConverter,
     MarketCapConverter,
+    IndustryBreakdownConverter,
+    IndustryGroupBreakdownConverter,
+    BondStatisticsConverter,
     MetaConverter
 } from './Converters';
 import {
```
jedrzejruta, line 34: Evil space :p
```suggestion
import {
```

src/Shared/SharedSecurityDetails.ts
```diff
@@ -100,10 +105,73 @@ export function initConverter (
...
     }
 }
+
+
+export const getBreakdown = (
+    id: string,
+    breakdown: SecurityDetailsJSON.GenericBreakdownType[],
+    table: External.DataTable,
+    colName: string,
+    hasMultiple: boolean
```
jedrzejruta, line 142: Another one :p
```suggestion
    hasMultiple: boolean
```

jedrzejruta (approved):
Not much to add, great amount of work! :tada:

src/Shared/Converters/BondStatisticsConverter.ts
```diff
@@ -0,0 +1,127 @@
...
+
+
+    public constructor (
+        options?: SecurityDetailsConverterOptions
+    ) {
+        super(options);
+
+        this.metadata = {
+            columns: {},
+            ...(options && options.hasMultiple && { ids: [] }),
```
kamil-musialowski, line 56: Change to optional chaining in all occurrences of that

```suggestion
            ...(options?.hasMultiple && { ids: [] }),
```

src/Shared/Converters/IndustryBreakdownConverter.ts
```diff
@@ -0,0 +1,125 @@
...
+
+
+    public constructor (
+        options?: SecurityDetailsConverterOptions
+    ) {
+        super(options);
+
+        this.metadata = {
+            columns: {},
+            ...(options && options.hasMultiple && { ids: [] }),
```
kamil-musialowski, line 57: Same here, optional chaining

kamil-musialowski (changes requested):
Nitpicking, looks good!

PR #90, opened 2025-05-12
Fixed inactive filter buttons in Investor Preferences demo

jedrzejruta:
Fixed inactive filter buttons after loading env credentials for Investor Preferences demo.


pawelfus (approved):
Thanks!

PR #98, opened 2025-06-03
Added BondStyleBoxBreakdowns converter.

kamil-musialowski:
Added BondStyleBoxBreakdowns converter.

TODO: 
- [ ] Add tests
- [ ] Check XRay compatibility
- [ ] Add Migration Package demo

### DIFF
--- a/src/SecurityDetails/SecurityDetailsOptions.ts
+++ b/src/SecurityDetails/SecurityDetailsOptions.ts
@@ -104,6 +104,7 @@ export type SecurityDetailsConverterType  = (
    | 'IndustryGroupBreakdown'
    | 'BondStatistics'
    | 'Meta'
+   | 'BondStyleBoxBreakdown'
  );
 
 
--- a/src/SecurityDetails/SecurityDetailsJSON.ts
+++ b/src/SecurityDetails/SecurityDetailsJSON.ts
@@ -60,6 +60,7 @@ namespace SecurityDetailsJSON {
     }
 
     export type PortfoliosType = {
+        BondStyleBoxBreakdown: BondStyleBoxBreakdownType[];
         AssetAllocations: AssetAllocationType[]
         RegionalExposure: RegionalExposureType[]
         GlobalStockSectorBreakdown: GlobalStockSectorBreakdownType[]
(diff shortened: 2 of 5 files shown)


jedrzejruta (commented):
Great work! 👏 

It would be best to align this converter with oncoming Security Details refactor: https://github.com/highcharts/connectors-morningstar/pull/95 (setting metadata is moved to the connector).

src/SecurityDetails/SecurityDetailsOptions.ts
```diff
@@ -104,6 +104,7 @@ export type SecurityDetailsConverterType  = (
    | 'IndustryGroupBreakdown'
    | 'BondStatistics'
    | 'Meta'
+   | 'BondStyleBoxBreakdown'
```
pawelfus, line 107: Just `BondStyleBox`. It is not a breakdown of grouped data, but rather the opposite: a grouping into specific categories.

pawelfus: Oh, just realized this is the name of the property in the returned response.. Nevermind then, compatibility with response is more important.

tests/SecurityDetails/SecurityDetailsConnector.test.ts
```diff
@@ -390,3 +390,29 @@ export async function MetaLoad (
...
+    await connector.load();
+
+    Assert.deepStrictEqual(
+        connector.table.getColumnNames(),
+        [
+            'BondStyleBoxBreakdown_Type',
+            'BondStyleBoxBreakdown_SalePosition_N'
+        ],
+        'Bond Style Box Breakdown table should exist of expected columns.'
+    );
```
pawelfus, line 417: Shall we check if ANY data is returned and parsed? I mean we know that there will be 9 categories, because the matrix is 3x3. I think we can ignore category=99.

pawelfus (approved):
Nice! See my comments below, nothing a showstopper.

tests/SecurityDetails/SecurityDetailsConnector.test.ts
kamil-musialowski: Hmm, how about checking if the length of the values array is `>=9` ?

src/SecurityDetails/SecurityDetailsOptions.ts
kamil-musialowski: Yeah, I would stick to the props returned from API. Especially, that we have `BondStyleBox` in XRay and they have a different structure (you know)

tests/SecurityDetails/SecurityDetailsConnector.test.ts
pawelfus: Sounds good, let's try it

DJTechnoo (approved):
Looks great! _Note: We can refactor using the `getBreakdown` method in the future. Waiting for API bugfix with triple salespositions + make `NotClassified` optional in the `getBreakdown`_

PR #101, opened 2025-06-18
Fixed failing tests.

pawelfus:
In breakdowns I have simply added the extra columns. This is not perfect solution - once the security change it's portfolio products, it might be broken again.

For RNA - replaced fixed dates with simply last 10 events.


kamil-musialowski (approved):
The solution with the columns should be good enough for now. It's better than having the failing tests on master. 👍

jedrzejruta (approved):

PR #105, opened 2025-07-02
Fixed empty columns in MetaConverter

DJTechnoo:
Fixed empty columns in MetaConverter using `SecurityCompare`.


kamil-musialowski (approved):
Nice!

PR #120, opened 2025-07-28
Added EquityStyle and FixedIncomeStyle converters to XRay US.

kamil-musialowski:
- [x] tests are not working
- [x] decide what to do with 'Unclassified' type
- [x] migration package demo

### DIFF
--- a/src/XRayUS/XRayUSJSON.ts
+++ b/src/XRayUS/XRayUSJSON.ts
@@ -56,7 +56,7 @@ namespace XRayUSJSON {
         AsOfDate: string;
         EquityStyleBreakdown: {
             AsOfDate: string;
-            Portfolio: GenericEquityStyleBreakdownItem;
+            Portfolio: EquityStyleBreakdownItem;
             SecurityBreakdown: Array<EquityStyleSecurityBreakdownItem>;
         };
     }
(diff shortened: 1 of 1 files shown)


demos/core-us-investment-style/demo.js
```diff
@@ -65,8 +65,86 @@ async function displayXRayUS (postmanJSON) {
...
+            align: 'right',
+            y: 75,
+            symbolRadius: 0,
+            itemMarginTop: 9,
+            itemMarginBottom: 9
+        },
+        tooltip: {
+            pointFormat: `
+                <b> {series.yAxis.categories.(point.y)}
+                    {series.xAxis.categories.(point.x)
```
jedrzejruta, line 129: Missing right-hand side closing bracket.
```suggestion
                    {series.xAxis.categories.(point.x)}
```

jedrzejruta (approved):
Nicely done! 👏

demos/core-us-investment-style/demo.html
```diff
@@ -18,13 +19,13 @@
         </style>
     </head>
     <body>
-        <h1>XRay US Connector</h1>
+        <h1>XRay US Connector Investment Style</h1>
```
pawelfus, line 22: Let's follow the convention in the title and here too - use the same what is in the menu on the left.

pawelfus (changes requested):
One tiny thing and the one already mentioned in the migration package: we need to support all breakdown levels (securities too). Add them to both parsers (+tests).

bm64 (approved):
LGTM 👍

demos/core-us-investment-style/demo.js
```diff
@@ -133,8 +133,10 @@ async function displayXRayUS (postmanJSON) {
             name: 'Portfolio Weight',
             borderWidth: 1,
             borderColor: '#e5e7e9',
-            keys: ['Type', 'value', '_', 'x', 'y'],
-            data: connector.dataTables.EquityStyle.getRows(0, 9),
+            keys: ['Type', 'value', 'x', 'y'],
+            data: connector.dataTables.EquityStyle.getRows(
+                void 0, void 0, ['Type', 'Value', 'Style', 'Size']
+            ),
```
pawelfus, line 139: Cool! But does it skip 10th category ("Unclassified")? Bottom left corner probably has two points

kamil-musialowski: I have double checked everything, but it seems to be absolutely fine 🤔 
Here is the output that we use to populate the stylebox, I have also compared it with original dataTable values and they are matching.

image

pawelfus: Okay, double-checked. It is fine, Unclassified has a separate column (XRay EU populates this here, category 99:
image
).

pawelfus (approved):
Thanks!

pawelfus (approved):
Just one thing, thanks!

PR #123, opened 2025-07-31
Added CreditQuality converter to XRay US.

jedrzejruta:
Added CreditQuality converter to XRay US.


kamil-musialowski:
I think it would be better to base this commit on this PR https://github.com/highcharts/connectors-morningstar/pull/120, to avoid many conflicts and unnecessary changes - could you merge it here? 🤔

pawelfus:
Note:
Test failing because of changed response's fields order. Fixed in: https://github.com/highcharts/connectors-morningstar/pull/124

pawelfus (approved):
Thanks!

DJTechnoo (approved):
LGTM! 🥳

src/XRayUS/XRayUSJSON.ts
```diff
@@ -42,8 +42,16 @@ namespace XRayUSJSON {
...
-        };
+            FixedIncomeAnalysis: FixedIncomeAnalysis;
+            InvestmentStyle: InvestmentStyle;
+        }
+    }
+
+    export interface FixedIncomeAnalysis {
+        CreditQuality: {
+            Portfolio: Record<string, number>;
+            Benchmark: Record<string, number>;
```
kamil-musialowski, line 53: Can we change the `Record` to more explicit types?

src/Shared/USConverters/CreditQualityConverter.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (benchmark) {
+            let rowIndex = 0;
+
+            for (const [key, value] of Object.entries(benchmark)) {
+                table.setCell('Type', rowIndex, key);
+                table.setCell('Value_Benchmark' + columnSuffix, rowIndex, value);
+
+                ++rowIndex;
+            }
+        }
```
kamil-musialowski, line 88: Once you change the types to more explicit ones instead of `Record` you won't be able to use Object.entires, since by default `value` is a type of any in TS, you can use the approach that I have used here https://github.com/highcharts/connectors-morningstar/pull/120/commits/9107ad20e0361f6dcbabe7b83b6273fc7cc5db83 instead.

kamil-musialowski (changes requested):
Sorry, more nitpicking 😬

bm64 (approved):

PR #125, opened 2025-07-31
feature/calendar-year-returns-converter

bm64:
Added CalendarYearReturnsConverter in Performance Connector.


src/Performance/Converters/CalendarYearReturnsConverter.ts
```diff
@@ -0,0 +1,101 @@
...
+            portfolioPerformance = userOptions.json,
+            calendarYearReturn = portfolioPerformance.Performance[0].Returns.CalendarYearReturn,
+            portfolioName = portfolioPerformance.Performance[0].PortfolioName,
+            columnSuffix = hasMultiple ? `_${portfolioName}` : '';
+
+        const calendarYearData = calendarYearReturn.Portfolio.CalendarYear;
+        const benchmarkData = calendarYearReturn.Benchmark?.CalendarYear;
+
+        for (let i = 0; i < calendarYearData.length; i++) {
+            const { Id, Value: PortfolioValue } = calendarYearData[i];
```
pawelfus, line 76: I think `PortfolioValue` -> `portfolioValue`. This is not a class. The same with other variables

src/Performance/PerformanceConnector.ts
```diff
@@ -57,7 +57,7 @@ export interface PerformanceConverter extends MorningstarConverter {
  * */
 
 export const DATA_TABLES = [
-    { key: 'TestConverter' }
+    { key: 'CalendarYearReturns' }
```
pawelfus, line 60: Should we just use `CalendarYearReturn` (singular) instead of plural form (`CalendarYearReturns`)? To match the response.

src/Performance/Converters/CalendarYearReturnsConverter.ts
```diff
@@ -0,0 +1,101 @@
...
+            const idColumn = 'Id';
+            const portfolioValueColumn = 'Portfolio_Value';
```
pawelfus, line 78: Thoughts about the name? @jedrzejruta @kamil-musialowski @DJTechnoo ?

src/Performance/Converters/CalendarYearReturnsConverter.ts
```diff
@@ -0,0 +1,101 @@
...
+    /* *
+     *
+     *  Functions
+     *
+     * */
+
+
+    public override parse (
+        options: PerformanceConverterOptions,
+        hasMultiple?: boolean
```
pawelfus, line 60: Doesn't work with multiple portfolios (only one is parsed):

Image

docs/connectors/morningstar/performance.md
```diff
@@ -16,7 +16,7 @@ Returned data depends on the selected view and config and may include tracking e
 
 Currently the following data points are supported in the Performance converter:
```
pawelfus, line 17: Connector*

pawelfus (changes requested):
Nice! A thew things, though.

src/Performance/Converters/CalendarYearReturnsConverter.ts
DJTechnoo: I think it is a good name, makes sense, short, familiar and not clashing with benchmark's values

DJTechnoo (approved):
It's a YES from me! 🎊

src/Performance/Converters/CalendarYearReturnsConverter.ts
pawelfus: It also supports multiple portfolios in one request, probably following naming convention from EU should be a good idea.

demos/core-us-calendar-year-return/demo.js
```diff
@@ -54,5 +52,40 @@ async function displaySecurityDetails (postmanJSON) {
...
+        },
+        tooltip: {
+            valueSuffix: '%'
+        },
+        series: [{
+            name: 'Portfolio',
+            data: connector.dataTables.CalendarYearReturn.getRows(
+                void 0,
+                void 0,
+                ['Id', 'Portfolio_Value']
```
pawelfus, line 78: Wrong column name

pawelfus (changes requested):

pawelfus (approved):
Thanks!

PR #126, opened 2025-08-04
Added Asset Allocation converter for XRay US.

kamil-musialowski:
- [x] migration package demo


docs/connectors/morningstar/x-ray.md
```diff
@@ -84,6 +84,7 @@ Currently the following data points are supported in the Americas X-Ray converte
 
 - **EquityStyle**
 - **FixedIncomeStyle**
+= **AssetAllocation**
```
pawelfus, line 87:
```suggestion
- **AssetAllocation**
```

src/Shared/USConverters/index.ts
```diff
@@ -36,6 +37,7 @@ import FixedIncomeStyleConverter from './FixedIncomeStyleConverter';
 
 export * from './EquityStyleConverter';
 export * from './FixedIncomeStyleConverter';
+export * from './USAssetAllocationsConverter';
```
pawelfus, line 40: Either use `US-` prefix in each converter, or simply let the converters be in `US-..` directory.

src/Shared/USConverters/USAssetAllocationsConverter.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (benchmark) {
+            setAssetAllocationCells(benchmark, '_Benchmark');
+        }
+
+        function setAssetAllocationCells (
+            asset: Array<XRayUSJSON.AssetAllocationItem>,
+            benchmarkSuffix = ''
+        ) {
+            for (let i = 0; i < asset.length; i++) {
+                table.setCell(assetAllocation.Id + '_Type' + columnSuffix, i, asset[i].Id);
```
pawelfus, line 82: Q: Do we need `Id` in the column name?
Are you able to receive two different asset allocation settings? Tested with:
```
        "AssetClassGroupConfigs": {
            "AssetClassGroupConfig": [
                {
                    "Id": "ACG-USDETAILED-15"
                }, 
                {
                    "Id": "ACG-USBROAD"
                }
            ]
        },
```

And returns only the first one.

src/Shared/USConverters/USAssetAllocationsConverter.ts
```diff
@@ -0,0 +1,99 @@
...
+                table.setCell('L' + benchmarkSuffix + columnSuffix, i, asset[i].Long);
+                table.setCell('S' + benchmarkSuffix + columnSuffix, i, asset[i].Short);
+                table.setCell('N' + benchmarkSuffix + columnSuffix, i, asset[i].Net);
+            }
```
pawelfus, line 86: Could we also parse `SecurityBreakdown` like we did in Style converters?

src/Shared/USConverters/USAssetAllocationsConverter.ts
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * */
+
+
+    public override parse (
+        options: XRayUSConverterOptions,
+        hasMultiple?: boolean
+    ): void {
+        const table = this.table,
+            columnSuffix = hasMultiple ? `_${options.json.PortfolioName}` : '',
```
pawelfus, line 64: Unnecessary duplicate of `type` in multi-XRay: 

Image

src/Shared/USConverters/USAssetAllocationsConverter.ts
pawelfus: Tested also XRay with two portfolios, only one asset class can be set.

pawelfus (changes requested):
Nicely done! Some minor changes needed.

src/Shared/USConverters/index.ts
kamil-musialowski: In that case, I have already created `USConverters` folder, so I'll remove the US from converter name and we're good to go 😄

jedrzejruta: Yes, separate folder for US is simpler

kamil-musialowski: Thanks guys, I consider it done for now - let's stick with existing `USConverters` and remove the `US` from the converter name.

docs/connectors/morningstar/x-ray.md
```diff
@@ -84,6 +84,7 @@ Currently the following data points are supported in the Americas X-Ray converte
 
 - **EquityStyle**
 - **FixedIncomeStyle**
+- **AssetAllocation**
```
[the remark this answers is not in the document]
kamil-musialowski: Although it's called `AssetAllocation` in both EU and US responses json, let's stick with `AssetAllocations` to avoid changing the names in EU code.

jedrzejruta (approved):
Clever way with the mixed order of SecurityBreakdown! 🤓

PR #127, opened 2025-08-05
Added FundStatistics converter to XRay US endpoint.

jedrzejruta:
TODO:
- [ ] With upcoming major Dashboard release, add PortfolioAnalyzed for portfolio and Analyzed and NotAnalyzed for securityBreakdown item to dataTable metadata
- [ ] Migration package demo

### DIFF
--- a/tests/XRayUS/XRayUS.test.ts
+++ b/tests/XRayUS/XRayUS.test.ts
@@ -128,8 +128,8 @@ export async function xRayUSConnectorLoad (
         [
             'Type',
             'Value',
-            'Value_FOUSA00DFS',
-            'Value_FOUSA00C3O'
+            'Value_FOUSA00C3O',
+            'Value_FOUSA00DFS'
         ],
         'FundStatistics table should return expected column names.'
     );
--- a/tests/XRayUS/XRayUS.test.ts
+++ b/tests/XRayUS/XRayUS.test.ts
@@ -122,4 +122,20 @@ export async function xRayUSConnectorLoad (
         connector.dataTables.FixedIncomeStyle.getRowCount() > 0,
         'FixedIncomeStyle connector should not return empty rows.'
     );
+
+    Assert.deepStrictEqual(
+        connector.dataTables.FundStatistics.getColumnNames(),
+        [
+            'Type',
+            'Value',
+            'Value_FOUSA00DFS',
+            'Value_FOUSA00C3O'
+        ],
+        'FundStatistics table should return expected column names.'
+    );
+
+    Assert.ok(
+        connector.dataTables.FundStatistics.getRowCount() > 0,
+        'FundStatistics table should not contain empty rows.'
+    );
 }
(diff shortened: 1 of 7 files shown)


pawelfus:
> [ ] With upcoming major Dashboard release, add PortfolioAnalyzed for portfolio and Analyzed and NotAnalyzed for securityBreakdown item to dataTable metadata

Please add this TODO item as comment in code, where it should be applied (I guess in parser)

tests/XRayUS/XRayUS.test.ts
```diff
@@ -136,4 +136,20 @@ export async function xRayUSConnectorLoad (
...
         'FixedIncomeStyle connector should not return empty rows.'
     );
+
+    Assert.deepStrictEqual(
+        connector.dataTables.FundStatistics.getColumnNames().sort(),
+        [
+            'Type',
+            'Value',
+            'Value_FOUSA00C3O',
+            'Value_FOUSA00DFS'
```
pawelfus, line 146: Q1: Should we generate empty columns for empty security items?

Image

Q2: Won't the test break, when they analyze more securities in the future, and list will be longer?

pawelfus (approved):
Looks good, two questions though.

tests/XRayUS/XRayUS.test.ts
jedrzejruta: @pawelfus These are both very valid questions:
1. I find it quite weird that some securities return an empty item, and some aren't listed at all depending on the request. I'd assume that all securities should be listed, and empty ones should have  a `NotAnalyzed: 100` property at all times. Since we don't have the dataTable metadata properties, we could create empty columns for the not analyzed security items.
2. I'm afraid it will break, so adding empty columns could be a foolproof fix.

pawelfus: 1. Yes, and no. Let's analyze the holdings in the request:
    - `AAPL` is Stock (`ST` in SecurityReference), not a Fund (`FO`, `FE`, `EX` etc), so impossible to have FundStatistics
    - the remaining four are:
      - 3x `FO` - funds, okay
      - 1x unknown, but listed anyway, okay (`0P00002NW8`)

bm64 (approved):
Nothing to add, looks good.

tests/XRayUS/XRayUS.test.ts
jedrzejruta: 1. Yes, and weirdly APPL has a record with `NotAnalyzed: 100`, even though it's a Stock and not a Fund, making it much more confusing 🤔

pawelfus: True, but it is not entirely incorrect. Probably simpler than another flag like "Analyzable".

pawelfus (changes requested):
Resolve conflicts please & implement empty securities

pawelfus (approved):
Thanks!

PR #128, opened 2025-08-05
feature/perf-trailing-returns-converter

bm64:
Added trailing returns converter.


jedrzejruta (approved):
LGTM! 🎉

pawelfus (changes requested):
See comments below.

In addition:
- resolve conflicts
- fix names and title (human readable format, correct the title):

Image

- Q: Do we two columns with IDs? The same for CalendarYear? According to docs, this is standard enum with specified categories. In CalendarYear it is always last 10 years:

Image

PR #131, opened 2025-08-08
Added GrowthConverter

DJTechnoo:
Added `GrowthConverter` to `HypoPerformanceConnector`.


src/HypoPerformance/HypoPerformanceJSON.ts
```diff
@@ -25,9 +25,31 @@
...
+
+    export interface PortfolioSummary {
+        AsOfDate: string;
+        PortfolioSummaryChart: PortfolioSummaryChart;
+    }
+
+    export interface PortfolioSummaryChart {
+        Frequency: string;
+        Portfolio: Portfolio;
+        Benchmark: Portfolio;
```
pawelfus, line 44: Missing `NetAmountInvested`

demos/hypo/demo.js
```diff
@@ -100,5 +100,43 @@ async function displaySecurityDetails (postmanJSON) {
 
     await connector.load();
 
+    Highcharts.chart('container', {
+        chart: {
+            type: 'column'
```
pawelfus, line 105: Let's use spline series

demos/hypo/demo.js
```diff
@@ -30,9 +30,9 @@ async function displaySecurityDetails (postmanJSON) {
                 capitalGainTaxRate: 0,
                 stateIncomeTaxRate: 0,
                 dividendTaxRate: 0,
-                illustrationTrailingTimePeriod: 'EarliestCommon',
-                startDate: '1998-12-31T00:00:00.000Z',
-                endDate: '2008-12-31T00:00:00.000Z',
+                illustrationTrailingTimePeriod: 'Customized',
+                startDate: '2002-01-01',
+                endDate: '2004-01-01',
```
pawelfus, line 35: Q: Why last points are always `0`? Can we do something about it? Tested different dates, and still the same problem. It looks strange on the line chart..

Image

demos/hypo/demo.js
```diff
@@ -30,9 +30,9 @@ async function displaySecurityDetails (postmanJSON) {
                 capitalGainTaxRate: 0,
                 stateIncomeTaxRate: 0,
                 dividendTaxRate: 0,
-                illustrationTrailingTimePeriod: 'EarliestCommon',
-                startDate: '1998-12-31T00:00:00.000Z',
-                endDate: '2008-12-31T00:00:00.000Z',
+                illustrationTrailingTimePeriod: 'Customized',
+                startDate: '2002-01-01',
```
pawelfus, line 34: How about we display ~5 years of data? Eg 2020 - 2025

pawelfus (changes requested):
Nice!

With addition to the inline comments:

1. I think we have a few redundant columns, when testing two portfolios (Date and Id):

Image

DJTechnoo:
Regarding the redundant Date and Id columns: They may not always be the same and might have different starting dates and different lengths per portfolio depending on the securities and the common dates between them.

kamil-musialowski (approved):
Looks nice to me! Good job :)

jedrzejruta (approved):
Great job, thanks! 👏

bm64 (approved):
Nothing to add, great job

demos/hypo/demo.js
pawelfus: Requested explanation from Morningstar, awaiting response..

pawelfus (commented):
On hold

PR #132, opened 2025-08-12
Added MPT Statistics converter to XRay US.

jedrzejruta:


kamil-musialowski (approved):
Short and simple, nice!

src/Shared/USConverters/MPTStatisticsConverter.ts
```diff
@@ -0,0 +1,94 @@
...
+            columnSuffix = hasMultiple ? `_${options.json.PortfolioName}` : '',
+            mptStatistics = options.json.Risks.MPTStatistics;
+
+        if (mptStatistics) {
+            let rowIndex = 0;
+
+            for (const breakdownItem of mptStatistics) {
+                const portfolio = breakdownItem.Portfolio;
+
+                table.setCell('TrailingTimePeriod', rowIndex, breakdownItem.TrailingTimePeriod);
```
kamil-musialowski, line 72: I would suggest moving it before the for...of loop, see: https://github.com/highcharts/connectors-morningstar/pull/133/commits/0291e2b6b2184c5b97e00cc3a391588b0e898604

kamil-musialowski (changes requested):

src/Shared/USConverters/MPTStatisticsConverter.ts
kamil-musialowski: Confirmed! ✅

DJTechnoo (approved):
LGTM!

PR #134, opened 2025-08-14
Added Matrix Correlation converter to XRay US connector.

jedrzejruta:
TODO: 
- [x] Migration package demo

### DIFF
--- a/src/Shared/USConverters/CorrelationMatrixConverter.ts
+++ b/src/Shared/USConverters/CorrelationMatrixConverter.ts
@@ -63,7 +63,7 @@ export class CorrelationMatrixConverter extends MorningstarConverter {
             columnSuffix = hasMultiple ? `_${options.json.PortfolioName}` : '',
             correlationMatrix = options.json.Risks.CorrelationMatrix;
 
-        if (correlationMatrix.length) {
+        if (correlationMatrix) {
             for (const correlationMatrixItem of correlationMatrix) {
                 const { TrailingTimePeriod, Correlations } = correlationMatrixItem;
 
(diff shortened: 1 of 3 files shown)


DJTechnoo (approved):

pawelfus (changes requested):
Format change request:

- from `[ID]_[extra]` to what we agreed: `[type/property]_[extra]_[security_id]` in other words: `Year3_FOUSA...` instead of `FOUSA..._Year3`

In addition, is it possible to add keys/indexes (like in StyleBox heatmaps)? So we could avoid this in demos:

Image

pawelfus:
Just discovered another issue, could you confirm the values in columns are correct? I'm afraid test should cover that (eg security with index = 0, has a value of 1 on index  = 0 etc). The response from API seems to be correct, tables:
Zrzut ekranu 2025-08-26 o 11 37 16
Year3, Portfolio 1:
Zrzut ekranu 2025-08-26 o 11 42 21

Year3, Portfolio 2:

Zrzut ekranu 2025-08-26 o 11 42 53

src/Shared/USConverters/CorrelationMatrixConverter.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    public override parse (
+        options: XRayUSConverterOptions,
+        hasMultiple?: boolean
+    ): void {
+        const table = this.table,
+            columnSuffix = hasMultiple ? `_${options.json.PortfolioName}` : '',
+            correlationMatrix = options.json.Risks.CorrelationMatrix;
+
+        if (correlationMatrix.length) {
```
pawelfus, line 66: That causes an error: 

Image

PR #137, opened 2025-08-17
feature/performance-risk-statistics

bm64:
Added Risk Statistics converter for Performance connector.

### DIFF
--- a/src/Performance/Converters/RiskStatisticsConverter.ts
+++ b/src/Performance/Converters/RiskStatisticsConverter.ts
@@ -87,10 +87,10 @@ export class RiskStatisticsConverter extends MorningstarConverter {
                 for (const security of Security) {
                     const { SecurityId, RiskStatisticsItem, Weight } = security;
 
-                    table.setCell(`${SecurityId}_Weight${columnSuffix}`, i, Weight);
+                    table.setCell(`Weight_${SecurityId}${columnSuffix}`, i, Weight);
 
                     for (const key of Object.keys(RiskStatisticsItem || {}) as Array<keyof typeof RiskStatisticsItem>) {
-                        table.setCell(`${SecurityId}_${key}${columnSuffix}`, i, RiskStatisticsItem[key]);
+                        table.setCell(`${key}_${SecurityId}${columnSuffix}`, i, RiskStatisticsItem[key]);
                     }
                 }
             });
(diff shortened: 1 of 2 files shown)


kamil-musialowski (approved):
Looks familiar 😄

jedrzejruta (approved):
In the future we might combine some converters for Performance and XRay US, let's have them separate for now.

Thanks!

DJTechnoo (approved):
LGTM!

src/Performance/Converters/RiskStatisticsConverter.ts
```diff
@@ -0,0 +1,108 @@
...
+                    Portfolio,
+                    Security,
+                    TrailingTimePeriod,
+                    DataFrequency
+                } = statistic;
+
+                table.setCell('TrailingTimePeriod', i, TrailingTimePeriod);
+                table.setCell('DataFrequency', i, DataFrequency);
+
+                for (const key of Object.keys(Benchmark) as Array<keyof typeof Benchmark>) {
```
pawelfus, line 79: Q: Can `benchmark` be undefined?

src/Performance/Converters/RiskStatisticsConverter.ts
```diff
@@ -0,0 +1,108 @@
...
+                for (const key of Object.keys(Portfolio) as Array<keyof typeof Portfolio>) {
+                    table.setCell(key + columnSuffix, i, Portfolio[key]);
+                }
+
+                for (const security of Security) {
+                    const { SecurityId, RiskStatisticsItem, Weight } = security;
+
+                    for (const key of Object.keys(RiskStatisticsItem || {}) as Array<keyof typeof RiskStatisticsItem>) {
+                        table.setCell(`${SecurityId}_${key}${columnSuffix}`, i, RiskStatisticsItem[key]);
+                        table.setCell(`${SecurityId}_Weight${columnSuffix}`, i, Weight);
```
pawelfus, line 92: Shouldn't we set this cell once per each security, not per stat?

src/Performance/Converters/RiskStatisticsConverter.ts
bm64: @pawelfus I don't think so. Even if benchmark is not included in the body it still returns benchmark.

PR #138, opened 2025-08-18
feature/performance-statistics-converter

bm64:
Addec MPT Statistics Converter to Performance connector.

### DIFF
--- a/src/Performance/PerformanceJSON.ts
+++ b/src/Performance/PerformanceJSON.ts
@@ -54,13 +54,23 @@ namespace PerformanceJSON {
     }
 
     interface RiskStatisticsPortfolio extends RiskStatisticsItem {
+        GrossMean?: number;
+        GrossSharpeRatio?: number;
+        GrossStandardDeviation?: number;
+        InformationRatio: number;
+        GrossInformationRatio?: number;
+        TrackingError: number;
+        GrossTrackingError: number;
+        SortinoRatio: number;
+        GrossSortinoRatio?: number;
         ExcessReturn: number;
+        GrossExcessReturn?: number;
+    }
+    interface RiskStatisticsBenchmark extends RiskStatisticsItem {
         InformationRatio: number;
         TrackingError: number;
         SortinoRatio: number;
-    }
-    interface RiskStatisticsBenchmark extends RiskStatisticsItem {
-        SortinoRatio: number;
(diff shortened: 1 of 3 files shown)


src/Performance/PerformanceJSON.ts
```diff
@@ -31,9 +31,44 @@ namespace PerformanceJSON {
...
         Returns: Returns;
+        Risks: {
+            RiskStatistics: Array<RiskStatistics>;
+            MPTStatistics: Array<MPTStatisticsBreakdownItem>;
+        };
+    }
+
+    interface RiskStatistics {
+        AsOfDate: string;
+        TrailingTimePeriod: 'Year1' | 'Year2' | 'Year3' | 'Year5' | 'Year10';
```
jedrzejruta, line 42: Why not use the external Type that's already added at the end of the file?

jedrzejruta (changes requested):
Looks good, minor thing to adjust!

docs/connectors/morningstar/performance.md
```diff
@@ -16,7 +16,8 @@ Returned data depends on the selected view and config and may include tracking e
 
 Currently the following data points are supported in the Performance connector:
 
-- **CalendarYearReturnConverter**
+- **CalendarYearReturn**
+- **RiskStatistics**
```
pawelfus, line 20: I think copilot was right here, missing `MPT` in the list?

src/Performance/PerformanceJSON.ts
```diff
@@ -51,6 +86,31 @@ namespace PerformanceJSON {
...
     }
 
+    interface MPTStatisticsBreakdownItem {
+        AsOfDate: string;
+        TrailingTimePeriod: TrailingTimePeriod;
+        DataFrequency: string;
+        Portfolio: MPTStatisticsPortfolio;
+    }
+
+    interface MPTStatisticsPortfolio {
```
pawelfus, line 96: Missing all `Gross*` props when `IncludeGrossNetReturns` is defined (the same problem as with RiskStats converter):

Image

pawelfus (changes requested):
Please target the base branch, not the main branch. So only the real changes are diffed, and we are mitigating risk of unexpected merge errors.

pawelfus (approved):
Good, thanks!

PR #141, opened 2025-08-28
Fixed wrong region detect, added default US endpoint URL for PAUS connectors.

jedrzejruta:
New regex for region detection based on this list https://gist.github.com/wpsmith/7604842 and official [RFC documentation](https://datatracker.ietf.org/doc/html/rfc5646).

_Internal note:_ Unable to test for correct US URL in PAUS tests, since all tests are setup to hit _emea_ endpoint. Not exactly sure how to approach this one 🤔 

---
  - https://app.asana.com/0/0/1211160184122289


pawelfus (approved):
Thanks!

kamil-musialowski (approved):
Wow, nice!

PR #142, opened 2025-09-03
Added `riskStatistics` converter to security details.

jedrzejruta:
TODO: 
- [x] Migration package demo
- [ ] <s>Rename XRay demo in migration package</s>

### DIFF
--- a/src/SecurityDetails/SecurityDetailsJSON.ts
+++ b/src/SecurityDetails/SecurityDetailsJSON.ts
@@ -71,6 +71,8 @@ namespace SecurityDetailsJSON {
         Alphas?: GenericRiskStatisticsResponse[];
         ArithmeticMeans?: GenericRiskStatisticsResponse[];
         Betas?: GenericRiskStatisticsResponse[];
+        BenchmarkCorrelations?: GenericRiskStatisticsResponse[];
+        Correlations?: GenericRiskStatisticsResponse[];
         InformationRatios?: GenericRiskStatisticsResponse[];
         MaxDrawdowns?: DrawdownRiskStatisticsResponse[];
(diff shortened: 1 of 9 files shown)


src/SecurityDetails/SecurityDetailsJSON.ts
```diff
@@ -58,6 +58,53 @@ namespace SecurityDetailsJSON {
         ActualManagementFee: number;
         FundAttributes?: FundAttributesType;
         HistoricalPerformanceSeries?: SecurityDetailsHistoricalPerformance[];
+        RiskStatistics?: RiskStatisticsItem[];
+    }
+
+    interface RiskStatisticsItem {
```
kamil-musialowski, line 64: Here we have `Rsquareds` and `RSquareds`, is it intentional?

kamil-musialowski (approved):
Looks like a solid work, nothing to add apart from one minor pick!

src/SecurityDetails/SecurityDetailsJSON.ts
jedrzejruta: Whoops, thanks for noticing - one time is enough 😄

src/SecurityDetails/SecurityDetailsJSON.ts
```diff
@@ -58,6 +58,52 @@ namespace SecurityDetailsJSON {
...
```
pawelfus, line 64: Missing `Correlations` and `BenchmarkCorrelations`:

Image

pawelfus (changes requested):
Nice! Minor fixes needed

DJTechnoo (approved):
LGTM!

PR #151, opened 2025-11-12
Added Direct Web Services connector.

kamil-musialowski:
Added Direct Web Services connector for new Morningstar API.

In order to use new credentials in tests, please add `DWS_MORNINGSTAR_USERNAME` and `DWS_MORNINGSTAR_PASSWORD` (do not delete the old ones) in the `.env` file.

- [x] figure out multi data tables for one request

### DIFF
--- a/src/DirectWebServices/InvestmentsConnector/InvestmentsConnector.ts
+++ b/src/DirectWebServices/InvestmentsConnector/InvestmentsConnector.ts
@@ -65,13 +65,12 @@ export class InvestmentsConnector extends DWSConnector {
     public constructor (
         options: InvestmentsOptions
     ) {
-        const convertersToUse = pickConverters(options.converters),
-            dataTablesToUse = dataTablesFromConverters(convertersToUse);
+        const convertersToUse = pickConverters(options.converters);
 
         options = {
             ...InvestmentsConnector.defaultOptions,
             ...options,
-            dataTables: dataTablesToUse
+            dataTables: dataTablesFromConverters(convertersToUse)
         };
 
         super(options);
@@ -123,9 +122,11 @@ export class InvestmentsConnector extends DWSConnector {
             if (this.convertersToUse.find(c => c.key === type)?.children) {
                 for (const table of converter.getTables()) {
                     this.dataTables[table.id].setColumns(table.getColumns());
+                    this.dataTables[table.id].metadata = table.metadata;
                 }
             } else {
                 this.dataTables[type].setColumns(converter.getTable().getColumns());
+                this.dataTables[type].metadata = converter.getTable().metadata;
             }
 
             this.metadata.rawResponses.push({ type, json });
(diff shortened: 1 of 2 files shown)


pawelfus (changes requested):
Quick peek:
- the `meta` of the connector is empty (the response(s) should be there)
- `applyTableModifiers()` missing, but I guess it's related to the next point:
- connector should grab multiple converters (endpoints equivalents) and fetch them and parse, instead of one-to-one relation (one connector = one endpoint is what we want to drop)

src/DirectWebServices/DWSOptions.ts
```diff
@@ -0,0 +1,63 @@
...
+    [key: string]: Response;
+}
+
+export interface DWSMetadata extends MorningstarMetadata {
+    rawResponses: Array<{ type: InvestmentsConverterType; json: unknown }>;
+}
+
+export interface DWSOptions extends MorningstarOptions {
+    languageId?: 'ENG' | 'SPA' | 'FRA' | 'DEU' | 'ITA' | 'JPN' | 'CHI' | 'ZHO' | 'KOR';
+
```
jedrzejruta: Empty line could be removed here
```suggestion
```

src/DirectWebServices/InvestmentsConnector/InvestmentsConnector.ts
```diff
@@ -0,0 +1,132 @@
...
+        this.requests = createRequests(this.dataTables, converters, security);
+
+        await super.load();
+
+        for (const responseObject of this.responses) {
+            const [type, response] =
+                Object.entries(responseObject)[0] as [InvestmentsConverterType, Response];
+
+            // TODO: Change type casting to a valid one
+            const json = await response?.json() as MorningstarConverterOptions;
```
jedrzejruta: The type of the response could be pulled from the converter's JSON namespace, similarly like here:
https://github.com/highcharts/connectors-morningstar/pull/153/files#diff-81d41a5014a02108043b0ed83ff3954069d89f845c55d9d7f4855611e7eca4e3

pawelfus (approved):
Nothing to add from me (as long, as Askel's solution in his PR on multidatatable <-> one connector is somehow implemented).

src/DirectWebServices/InvestmentsConnector/InvestmentsConnector.ts
kamil-musialowski: After a discussion, we decided that it doesn't make sense and we should stick to `unknown` for now.

src/DirectWebServices/InvestmentsConnector/InvestmentsConnector.ts
```diff
@@ -113,7 +120,13 @@ export class InvestmentsConnector extends DWSConnector {
 
             converter.parse({ json });
 
-            this.dataTables[type].setColumns(converter.getTable().getColumns());
+            if (this.convertersToUse.find(c => c.key === type)?.children) {
+                for (const table of converter.getTables()) {
+                    this.dataTables[table.id].setColumns(table.getColumns());
```
PaulDalek: As an addition to this I would also add the metadata from the specific tables: `this.dataTables[table.id].metadata = table.metadata;`

pawelfus (approved):
Looks good! Please improve what Paweł suggested.

PR #152, opened 2025-11-12
adjustment/new-api-build

PaulDalek:
Prepared scripts building process for the new API.

---
  - https://app.asana.com/0/0/1211876468208348


pawelfus (changes requested):
Just one thing, the naming changes. Other than that looks good!

pawelfus (approved):
Thanks!

PR #154, opened 2025-11-20
Added FundStatistics metadata properties (Analyzed/NotAnalyzed).

jedrzejruta:
Added Analyzed/NotAnalyzed properties to the XRayUS FundStatistics metadata.

This is a POC that could be later reused in other converters, so that any information that does not fit into more generic `connector.metadata` can be accessed directly from the `dataTable.metadata` instead.

### DIFF
--- a/tests/Screeners/InvestmentScreener.test.ts
+++ b/tests/Screeners/InvestmentScreener.test.ts
@@ -51,10 +51,9 @@ export async function investmentScreenerLoad (
         'Connector table should exist of expected columns.'
     );
 
-    Assert.strictEqual(
-        connector.getTable().getRowCount(),
-        15,
-        'Connector table should have 15 rows.'
+    Assert.ok(
+        connector.getTable().getRowCount() > 0,
+        'Connector table should not contain empty rows.'
     );
 
     Assert.deepStrictEqual(


DJTechnoo (approved):
LGTM! 🎊

pawelfus (approved):
Thanks!

src/Shared/USConverters/FundStatisticsConverter.ts
```diff
@@ -78,24 +80,33 @@ export class FundStatisticsConverter extends MorningstarConverter {
...
+                    securityId = security.SecurityId;
                 let rowIndex = 0;
 
                 // Set empty columns for all securities
-                table.setColumn('Value_' + security.SecurityId);
+                table.setColumn('Value_' + securityId);
+
+                // Set metadata props for each security
+                metadata[`Analyzed_${securityId}`] = security.Analyzed;
+                metadata[`NotAnalyzed_${securityId}`] = security.NotAnalyzed;
```
kamil-musialowski: Can we go even further with consts and do something like this?

`const { FundStatisticsItem, SecurityId, Analyzed, NotAnalyzed } = security;`

kamil-musialowski (changes requested):
Just one suggestion, other than that it looks good!

PR #164, opened 2026-01-08
Improved column names in Investments Connector.

kamil-musialowski:


pawelfus (changes requested):
Missing changes in docs

PR #177, opened 2026-04-16
Added logic for replacing CDN urls with local files.

kamil-musialowski:
Run ```npm install``` + ```npm run demos```.

To see what script (and where from) is fetched, open DevTools and go the the Network tab.


PaulDalek (approved):
Good job! Other than some failing tests (probably because of some Morningstar changes again), everything looks ok.

DJTechnoo (approved):
LGTM! 🎉

package.json
```diff
@@ -22,6 +22,9 @@
         }
     },
     "devDependencies": {
+        "@highcharts/dashboards": "^4.1.0",
+        "@highcharts/grid-pro": "^2.2.0",
+        "highcharts": "^12.5.0",
```
jedrzejruta, line 27: Perhaps we could bump to 12.6.0, so that it's already on the latest version?

jedrzejruta (approved):
LGTM! 🎉

package.json
kamil-musialowski: @pawelfus

pawelfus: Not needed now

pawelfus (approved):
Thanks!

PR #181, opened 2026-04-27
[DO NOT MERGE] Adjusted demos to support new DataTable.

jedrzejruta:
Adjusted demos to support Highcharts v13 DataTable.

For testing, demos have a link to the dataTable PR, once it's approved the link should be reverted back to the original.

---
  - https://app.asana.com/0/0/1213701228416344

### DIFF
--- a/demos/core-us-trailing-returns/demo.js
+++ b/demos/core-us-trailing-returns/demo.js
@@ -66,6 +66,7 @@ async function displayPerformance (postmanJSON) {
 
     const dataTable = connector.getTable('TrailingReturns');
 
+    // Add column with name translation
     dataTable.setColumn(
         'Name',
         dataTable.getColumn('Id').map(id => periods[id] || id)
--- a/demos/core-us-trailing-returns/demo.js
+++ b/demos/core-us-trailing-returns/demo.js
@@ -64,19 +64,15 @@ async function displayPerformance (postmanJSON) {
         SinceInception: 'Since Inception'
     };
 
-    const portfolioData = connector.dataTables.TrailingReturns.getRows(
-        void 0,
-        void 0,
-        ['Id', 'Value']
-    ).map(([period, value]) => [periods[period], value]);
+    const dataTable = connector.getTable('TrailingReturns');
 
-    const benchmarkData = connector.dataTables.TrailingReturns.getRows(
-        void 0,
-        void 0,
-        ['Id', 'Value_Benchmark']
-    ).map(([period, value]) => [periods[period], value]);
+    dataTable.setColumn(
+        'Name',
+        dataTable.getColumn('Id').map(id => periods[id] || id)
+    );
 
     Highcharts.chart('container', {
+        dataTable,
         chart: {
             type: 'column'
         },
@@ -97,11 +93,17 @@ async function displayPerformance (postmanJSON) {
         series: [
             {
                 name: 'Portfolio',
-                data: portfolioData
+                dataMapping: {
+                    name: 'Name',
+                    y: 'Value'
+                }
             },
             {
                 name: 'Benchmark',
-                data: benchmarkData
+                dataMapping: {
+                    name: 'Name',
+                    y: 'Value_Benchmark'
+                }
             }
         ]
     });
--- a/demos/dws-equity-aggregates-residual-risk/demo.js
+++ b/demos/dws-equity-aggregates-residual-risk/demo.js
@@ -23,6 +23,7 @@ async function displayEquityAggregatesResidualRisk (postmanJSON) {
 
     // Set global options
     Highcharts.setOptions({
+        dataTable: connector.getTable('EquityAggregatesResidualRisk'),
         chart: {
             type: 'column'
         },
@@ -35,11 +36,16 @@ async function displayEquityAggregatesResidualRisk (postmanJSON) {
         },
         xAxis: {
             type: 'category'
+        },
+        plotOptions: {
+            series: {
+                dataMapping: {
+                    name: 'Type'
+                }
+            }
         }
     });
 
-    // Get data table
-    const dataTable = connector.getTable('EquityAggregatesResidualRisk');
     // TO DO: Uncomment alpha and beta series when query params are supported
 
     // Create chart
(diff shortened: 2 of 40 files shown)


demos/core-us-trailing-returns/demo.js
```diff
@@ -64,19 +64,15 @@ async function displayPerformance (postmanJSON) {
...
 
-    const benchmarkData = connector.dataTables.TrailingReturns.getRows(
-        void 0,
-        void 0,
-        ['Id', 'Value_Benchmark']
-    ).map(([period, value]) => [periods[period], value]);
+    dataTable.setColumn(
+        'Name',
+        dataTable.getColumn('Id').map(id => periods[id] || id)
+    );
```
jedrzejruta, line 73: For now I've come up with such way of adding name translation to the demos (instead of having to manually format tooltip and dataLabels), however I'm open to your input on this matter.

@pawelfus @kamil-musialowski @DJTechnoo @PaulDalek

pawelfus: Sounds good. Could add a comment above? Something like "// Add column with translation"

demos/dws-equity-residual-risk/demo.js
```diff
@@ -58,23 +62,22 @@ async function displayEquityResidualRisk (postmanJSON) {
         },
         series: [{
             name: 'Alpha',
-            data: riskDailyTable.getRows(
-                void 0,
-                void 0,
-                ['Type', 'Alpha']
-            )
+            dataMapping: {
+                name: 'Type',
```
pawelfus: Already defined above in `setOptions`

demos/core-hypothetical-performance/demo.js
```diff
@@ -100,7 +100,15 @@ async function displayHypoGrowth (postmanJSON) {
 
     await connector.load();
 
+    const dataTable = connector.getTable('Growth');
+
+    // A known issue with trailing 0's in the data
```
pawelfus, line 105: Q: Is this the only demo with such problem?

demos/core-security-details-asset-allocations/demo.js
```diff
@@ -28,10 +28,12 @@ async function displayAssetAllocations (postmanJSON) {
         '99': 'Unclassified'
     };
 
-    const chartData = connector.dataTables.AssetAllocations.getRowObjects().map(item => ({
-        name: typeMapping[item.MorningstarEUR3_Type],
-        y: item.MorningstarEUR3_N
-    }));
+    const dataTable = connector.getTable('AssetAllocations');
+
+    dataTable.setColumn(
```
pawelfus, line 34: Add a comment why it's needed

pawelfus (changes requested):
Looks really good, could you check a few things before accepting? See below:

demos/core-us-trailing-returns/demo.js
kamil-musialowski: Definitely better than manual format, good idea!

demos/dws-equity-sectors-breakdown/demo.js
kamil-musialowski: If you move this above setOptions, then you could change dataMapping.name to `type` instead of `'Type'`

kamil-musialowski (approved):
Good! Only one minor suggestion, the rest is fine ✅

demos/dws-fixed-income-sectors-breakdown/demo.js
```diff
@@ -49,99 +49,96 @@ async function displayFixedIncomeSectorsBreakdown (postmanJSON) {
...
-    // Get data table
-    const superSectorsBrkTable = connector.getTable('IncBrkSuperSectors');
-
     // Create chart
     Highcharts.chart('container-brk-super-sectors', {
+        dataTable: connector.getTable('IncBrkSuperSectors'),
+        plotOptions: {
+            series: {
+                dataMapping: {
+                    name: brkType
```
PaulDalek: All of the `name: brkType` occurances can be removed and moved to the `setOptions`, just like in some other file/example above.

demos/dws-equity-aggregates-residual-risk/demo.js
```diff
@@ -35,11 +36,16 @@ async function displayEquityAggregatesResidualRisk (postmanJSON) {
...
+                dataMapping: {
+                    name: 'Type'
+                }
+            }
         }
     });
 
-    // Get data table
-    const dataTable = connector.getTable('EquityAggregatesResidualRisk');
     // TO DO: Uncomment alpha and beta series when query params are supported
```
PaulDalek: I know this is outside of the scope of this PR but the `alpha` and `beta` series can be safely uncommented now as, from what I've checked, they are available again.

PaulDalek (changes requested):
Nice work! Only two suggestions from me.

demos/core-hypothetical-performance/demo.js
jedrzejruta: This is an ongoing issue with the response for Hypo Growth data, and besides this demo, we have one in migration package (growth-chart-hypo) with the same workaround.

pawelfus: Okay, thanks for the confirmation!

PaulDalek (approved):
Great, thanks!

pawelfus (approved):
Thanks!

All good, we will need to revert the script src's once v13 is released.

PS: The `npm run demos` should serve local files, when dependencies are installed. So the change could be actually in `bin/Library/Server.js` -> `CDNToLocal()`

DJTechnoo (approved):
Amazing stuff! 🎉

demos/stock-ohlcv-time-series/demo.js
pawelfus, line 1: This demo doesn't work (maximum call stack err)

demos/stock-rolling-return-time-series/demo.js
pawelfus, line 1: This demo doesn't work (maximum call stack err)

demos/stock-time-series/demo.js
pawelfus, line 1: This demo doesn't work (maximum call stack err)

demos/stock-security-compare/demo.js
pawelfus, line 1: This demo doesn't work (Uncaught (in promise) ne: Not Found)

demos/dws-stock-time-series/demo.js
```diff
@@ -26,6 +26,7 @@ async function initializeChart (postmanJSON) {
     await growthConnector.load();
 
     Highcharts.stockChart('container', {
+        dataTable: growthConnector.getTable(),
```
pawelfus, line 29: This demos is empty now, could you check it please? Seems unrelated to the changes, please investigate. Maybe there's workaround we can use.

Image

pawelfus (changes requested):
Thanks!

One last request before merging.

demos/dws-stock-time-series/demo.js
jedrzejruta: Looks like it fixed itself somehow, perhaps there was some issue on the backend as for some time it only returned data for one security, even though there were more securities in the request.

pawelfus (approved):

PR #182, opened 2026-04-30
enhancement/fixed-income-sectors-parents

PaulDalek:
Introduced a new column with the name of a parent sector.

### DIFF
--- a/src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
+++ b/src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
@@ -192,10 +192,12 @@ export class FixedIncomeSectorsBreakdownConverter extends MorningstarConverter {
                             fixedIncomeBreakdownMap.get(typeAndName) :
                             fixedIncomeMap.get(typeAndName);
 
-                        if (parent) {
-                            // Set parent sector value
-                            table.setCell(`${column}_Parent`, index, parent);
-                        }
+                        // Set parent sector value
+                        table.setCell(
+                            `${column}_Parent`,
+                            index,
+                            parent ?? 'UncategorizedSector'
+                        );
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


DJTechnoo (approved):
Looking good!

kamil-musialowski (changes requested):
Good one! Please also add that column in relevant test 👍

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
```diff
@@ -181,6 +185,18 @@ export class FixedIncomeSectorsBreakdownConverter extends MorningstarConverter {
...
                         table.setCell(`${column}_${match[3]}`, index, Number(value));
                     }
+
+                    // Get the parent sector only for Primary and Secondary
+                    if (!type.includes('SuperSector')) {
+                        const parent = field === 'fixdInc' ?
+                            fixedIncomeBreakdownMap.get(typeAndName) :
+                            fixedIncomeMap.get(typeAndName);
+
+                        if (parent) {
```
pawelfus: Q: What do you think about pushing sectors without parent match (eg new ones) to `UncategorizedSector`?

pawelfus (commented):
Nice, just one topic to discuss.

PaulDalek:
@kamil-musialowski Thanks for reminding, test file updated.

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
PaulDalek: Yes, that's a good idea, especially in situations where new sectors are added silently (which as we all know, can happen 😄). Updated logic, thanks!

kamil-musialowski (approved):
Thx! 👏

jedrzejruta (approved):
Magnificent, thanks! 🎉

pawelfus (approved):
Thanks!

PR #183, opened 2026-05-04
enhancement/merge-date-column

DJTechnoo:
Merged `Date` columns from `TimeSeries` into one shared column when multiple securities are fetched.

_The solution for this involves iterating through all date columns to verify the value indices match the dates in case of any holes in the dates for one or more securities._

### DIFF
--- a/src/DirectWebServices/TimeSeriesConnector/TimeSeriesConverter.ts
+++ b/src/DirectWebServices/TimeSeriesConnector/TimeSeriesConverter.ts
@@ -26,6 +26,7 @@
 import MorningstarConverter from '../../Shared/MorningstarConverter';
 import type { TimeSeriesConverterOptions } from './TimeSeriesOptions';
 import type { TimeSeriesResponse } from './TimeSeriesJSON';
+import { defined } from 'highcharts';
 
 
 /* *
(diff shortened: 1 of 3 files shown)


pawelfus (changes requested):
Solution looks great!

1. Good idea with adding another one to the demo! However in such case, we should switch to line series - readability reasons. 

2. Could you add name to both series?

3. Could you update Title?

src/DirectWebServices/TimeSeriesConnector/TimeSeriesConverter.ts
```diff
@@ -65,26 +66,41 @@ export class TimeSeriesConverter extends MorningstarConverter {
 
         if (investments) {
             const hasMultiple = investments.length > 1;
+
+            // Create a table of dates shared among all investments.
+            const dates = new Set<string>();
+
+            for (const investment of investments) {
```
kamil-musialowski, line 75: 1st loop

src/DirectWebServices/TimeSeriesConnector/TimeSeriesConverter.ts
```diff
@@ -65,26 +66,41 @@ export class TimeSeriesConverter extends MorningstarConverter {
...
+            }
+
+            const sortedDates = [...dates].sort();
+
+            table.setColumn(
+                'Date', sortedDates
+            );
+
+            // Assign point values via date indices.
             for (const investment of investments) {
```
kamil-musialowski: 2nd loop

kamil-musialowski (changes requested):
Good one! A performance question from me, now we are looping over `investments` twice. Can your logic be injected into the 2nd (old) loop? 🤔

jedrzejruta (approved):
Apart from comments from other reviewers I don't have much too add, nicely done! 🎉

src/DirectWebServices/TimeSeriesConnector/TimeSeriesConverter.ts
```diff
@@ -65,26 +66,41 @@ export class TimeSeriesConverter extends MorningstarConverter {
...
-                            `Value${columnSuffix}`,
-                            rowIndex,
-                            point.value
-                        );
-
-                        rowIndex++;
+
+                        const dateRowIndex = table.getRowIndexBy('Date', point.date);
+
+                        if (defined(dateRowIndex)) {
```
PaulDalek: Question: Is it possible that the `dateRowIndex` won't exist? If so, what should happen with this point? In this scenario it doesn't get saved at all.

PaulDalek (commented):
Good job! Other what was already mentioned, I don't have additional comments, except one question below.

kamil-musialowski (approved):
Great! Thank you ✋

kamil-musialowski:
@DJTechnoo You can merge master now, that should fix the failing tests.

pawelfus (approved):
Thanks!

PaulDalek (approved):
Great, thanks!

PR #184, opened 2026-05-05
enhancement/fixed-income-sectors-all

PaulDalek:
Added a new data table that contains all sector values, regardless of type (Super, Primary and Secondary).


jedrzejruta (approved):
Tested locally, the paths for all the tables seem correct, thanks! 👏

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
```diff
@@ -180,27 +197,70 @@ export class FixedIncomeSectorsBreakdownConverter extends MorningstarConverter {
...
+                const superArray = superColumns[key] as Array<string | null>,
+                    allArray = allColumns[key] as Array<string | null>,
+                    values =
+                        ['Fixed_Income_Path', 'Fixed_Income_Type']
+                            .includes(key) ?
+                            ['GovernmentPerCountry', 'Uncategorized'] :
+                            [null, null];
+
+                superArray.splice(index, 0, ...values);
+                allArray.splice(index, 0, ...values);
```
pawelfus: I think this will cause problems: `superColumns` and `allArray` are our column in dataTable, so should be updated via `setRow()` etc API

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
```diff
@@ -180,27 +197,70 @@ export class FixedIncomeSectorsBreakdownConverter extends MorningstarConverter {
...
-                            parent ?? 'UncategorizedSector'
-                        );
-                    }
+                    // Get the path of all sectors
+                    const path = field === 'fixdInc' ?
+                            fixedIncomeBreakdownPathMap.get(typeAndName) :
+                            fixedIncomePathMap.get(typeAndName),
+                        columnId = `${column}_Path`,
+                        cellValue = (path ?? ['Uncategorized', name].join('/'))
+                            .replace('Breakdown', '');
```
pawelfus: Q: Do we want to replace only the firet occurence?

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
```diff
@@ -180,27 +197,70 @@ export class FixedIncomeSectorsBreakdownConverter extends MorningstarConverter {
...
+                    allTable.setCell(columnId, allIndex, cellValue);
                 }
             }
+
+            // Supply missing sectors for uncategorized and country government
+            const superColumns = tablesObj['SuperSector'].columns,
+                allColumns = tablesObj['AllSector'].columns,
+                paths = superColumns['Fixed_Income_Path'];
+
+            // Find the last index of super sector before government per country
```
pawelfus: Q: Last or first?

pawelfus: Q2: Can `paths` be undefined?

pawelfus (changes requested):
Only one problem (with slice and mutating data), others are just questions.

kamil-musialowski (approved):
lgtm!

DJTechnoo (approved):
LGTM! 🎉

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
PaulDalek: In this specific case (the fixed income breakdown data), paths contain only one occurance of the Breakdown, so it's safe.

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
PaulDalek: 1. Yes, last. For the fixed income data, there are 6 main super sectors (`Government`, `Municipal`, `Corporate`, `Securitized`, `CashAndEquivalents` and `Derivative`). However, the data returned from the API categorizes the `Government` sector by country as separate super sectors (e.g. `fixedIncSuperSectorGovernmentPoland...`). There are 68 such sectors. Therefore, I decided to add another super sector, `GovernmentPerCountry`, after the main 6 and before the country related `Government` super sectors, and place all of them there. This makes the grid demo more readable by having a single category instead of 68 separate ones.

2. Yes, it might happen, for example when new sectors are added to the API response data and are not yet covered by the map object. I added this check (placing such sectors to `Uncategorized`) following your suggestion from the previous PR :).

src/DirectWebServices/Sectors/FixedIncomeSectorsBreakdownConverter.ts
PaulDalek: Ok, I’ve checked it once again, and it looks like the order does not really seem to be relevant here, so I changed it to apply the additional helper sector values at the end of each table.

By the way, `setRow()` with the `insert` parameter set to `true` seems to work incorrectly. It replaces the value instead of inserting it. Should it work that way? The comment says: `Whether to insert the row at the given index, or to overwrite the row`.

pawelfus (commented):
TBD: Gov by Country/Region, awaiting feedback from Grid team

PaulDalek:
@pawelfus The Government Per Country data is moved to a separate data table now.

pawelfus (changes requested):
Needed some correction in the hierarchy:


> 1.  Misplaced primaries / secondaries:
>    - `PrimarySectorCoveredBond` should live under Securitized, not Corporate.
>    - `Interest Rate Derivative` & `Treasury Future` belong under Government Related, not Derivative.
>    - `MBS`, `GSE/Agency`, `ABS`, `TD/CD`, `DerivativeCashOffsets`, `FRN`, `Municipal` belong under `Cash & Equivalents`, not in `Asset Backed` / `Agency MBS` / `Uncategorized`.
>    - `Bond Unit` belongs under `Future/Forward`, not `Cash`.
> 2. Outright missing secondaries: `401040 Unspecified`, plus the `Cash` entries (`Corporate`, `Government`) that have no other home in the TS map.

PaulDalek:
@pawelfus I was able to match all secondary sectors to the correct primary ones based on the documentation provided by MS. I briefly checked the sums of the values, and they now seem to be correct, at least for `fixedIncome`. Even though the `fixedIncomeBreakdown` values are very similar, there are 3 or 4 additional sectors.

pawelfus (changes requested):
Agreed on: separate tables for each Country-related secondary sector

PaulDalek:
@pawelfus Enhancements and corrections implemented. Also, data is correct (we checked it on the grid tree view demo).

pawelfus (approved):
Thanks!

PR #185, opened 2026-05-07
Added support for using pre-fetched JSON in Connectors.

kamil-musialowski:
Added `api.json` option to all Morningstar connectors, allowing users to bypass authentication and network requests by passing a pre-fetched JSON payload directly.


kamil-musialowski:
Failing tests are unrelated to this PR.

docs/connectors/morningstar.md
```diff
@@ -94,3 +94,37 @@ Example:
 ## Architecture
 This is a visualization of the Highcharts Morningstar Data Connector:
 ![Highcharts Morningstar Data Connector Architecture](architecture.png)
+
+## Offline mode with `api.json`
```
pawelfus: Could you also add example for new connectors (DWS)? Especially the one with multiple responses?

Nitpicking: use full name "`api.json` option", instead of `api.json`. First I thought it's broken URL.

src/Shared/MorningstarAPI.ts
```diff
@@ -214,6 +223,55 @@ export class MorningstarAPI {
         return response;
     }
 
+    protected fetchOffline (
```
pawelfus: Do we really need to go through this? All these overrides, stringifying and parsing back? 

Additional question: can we bypass the whole fetch logic and provide the JSON directly to converters?

src/Shared/MorningstarAPI.ts
```diff
@@ -131,13 +135,18 @@ export class MorningstarAPI {
...
-        requestInit: RequestInit = {}
+        requestInit: RequestInit = {},
+        offlineContext: MorningstarOfflineContext = {}
     ): Promise<Response> {
 
         if (url.searchParams.get('outputType') !== 'compactjson') {
             url.searchParams.set('outputType', 'json');
         }
 
+        if (this.options.json) {
```
pawelfus: Nitpicking: This should be before lines 142-144

pawelfus (changes requested):
Nice work! A few problems, see inline. And one overall: I think it shouldn't be called "offline" mode/JSON. How about Pre-fetched JSON?

demos/hypo-offline-json/demo.js
```diff
@@ -0,0 +1,71 @@
...
+                ['Date', 'Value_Benchmark']
+            ).slice(0, -1) // A known issue with trailing 0's in the data
+        }, {
+            name: 'Net Invested',
+            color: 'green',
+            data: connector.dataTables.Growth.getRows(
+                void 0,
+                void 0,
+                ['Date', 'Value_NetAmountInvested']
+            ).slice(0, -1) // A known issue with trailing 0's in the data
```
jedrzejruta: Instead of accessing directly from `connector.dataTables`, let's set a variable for the Growth table and fetch it once using `connector.getTable(...)`.

jedrzejruta (changes requested):
Nicely done, thanks! 🎉 

Minor thing to adjust below.

src/Shared/MorningstarAPI.ts
kamil-musialowski: 1. The logic is taken from `fetch()`  to keep the JSON available in response, and keep it readable more than once.

2. We could, but that would require modifying all connectors logic which now expect `response.json` instead of pure JSON.

kamil-musialowski:
@pawelfus 
> I think it shouldn't be called "offline" mode/JSON. How about Pre-fetched JSON?

I wasn't sure about the offline name either. But pre-fetched doesn't sound good to me. How about `localJSON`?

jedrzejruta (approved):
Thanks!

src/Shared/MorningstarAPI.ts
pawelfus: 1. I might be missing something - in which response? Why we have to pretend it's a response, while it can be just object `json` option should bypass the whole fetch logic. Similar to the way it works right now in JSON connector in Dashboards.
2. I kind of expected changes in all connectors (mostly because all of them repeat the same part):
```ts
        const json = await response.json() as unknown;

        this.converter.parse({ json });

        this.getTable().deleteColumns();
        this.getTable().setColumns(this.converter.getTable().getColumns());

        return this.applyTableModifiers();
```
Which could be extracted to MorningstarConnector's class. And unique implementations (eg DWS and SecurityDetails) should override default method. Something like this (pseudecode):
```ts
if (options.json) {
  return this.parseJSON(options.json);
}

const api = this.api = this.api || new MorningstarAPI(options.api);
...
const json = await response.json() as unknown;
return parseJSON(json);
```

Where:
```ts
parseJSON (json): Promise<this> {
  if (!this.converter) {
    // throw err?
  }
  this.converter.parse({ json });

  this.getTable().deleteColumns();
  this.getTable().setColumns(converter.getTable().getColumns());

  return this.applyTableModifiers();
}

**Review by pawelfus — COMMENTED** [review]
> @pawelfus
> 
> > I think it shouldn't be called "offline" mode/JSON. How about Pre-fetched JSON?
> 
> I wasn't sure about the offline name either. But pre-fetched doesn't sound good to me. How about `localJSON`?

Or just "json option"/"api.json option". Let's wait with naming discussion until we agree on solution first. Local/Offline names are both ambiguous. Tried to get some inspiration from Dashboards - [docs](https://www.highcharts.com/docs/dashboards/data-pool-and-connectors#jsonconnector-example) and [api ref](https://api.highcharts.com/dashboards/#interfaces/Data_Connectors_JSONConnectorOptions.JSONConnectorOptions) - it is simply called `JSONConnector`

**Review by PaulDalek — APPROVED** [review]
Great work implementing this idea! Other than maybe rethinking the response vs. providing raw object approach, I don’t have much to add.

**Review by DJTechnoo — APPROVED** [review]
Magnificent!

**Comment by kamil-musialowski** [comment]
RNA News test fails on master as well.

File: demos/dws-offline-json-asset-alloc-breakdown/demo.html [hunk 5]
```diff
@@ -0,0 +1,23 @@
...
+        <script src="https://code.highcharts.com/connectors/morningstar/connectors-morningstar-dws.js"></script>
+        <title>Highcharts Core + Morningstar Asset Allocation Breakdown (Offline JSON)</title>
+        <style>
+            body {
+                font-family: sans-serif;
+            }
+        </style>
+    </head>
+    <body>
+        <h1>Highcharts Core + Morningstar Asset Allocation Breakdown (Offline JSON)</h1>
```
pawelfus: Rename

demos/dws-offline-json-asset-alloc-breakdown/demo.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <script src="https://code.highcharts.com/highcharts.js"></script>
+        <script src="https://code.highcharts.com/connectors/morningstar/connectors-morningstar-dws.js"></script>
+        <title>Highcharts Core + Morningstar Asset Allocation Breakdown (Offline JSON)</title>
```
pawelfus: Rename

demos/hypo-offline-json/demo.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <script src="https://code.highcharts.com/highcharts.js"></script>
+        <script src="https://code.highcharts.com/connectors/morningstar/connectors-morningstar.js"></script>
+        <title>Highcharts Core + Morningstar Hypothetical Performance (Offline JSON)</title>
```
pawelfus: Rename

demos/hypo-offline-json/demo.html
```diff
@@ -0,0 +1,19 @@
...
+        <style>
+            body {
+                font-family: sans-serif;
+            }
+        </style>
+    </head>
+    <body>
+        <h1>Highcharts Core + Morningstar Hypothetical Performance (Offline JSON)</h1>
```
pawelfus: Rename

demos/index.html
```diff
@@ -114,6 +114,12 @@ <h3>Demos</h3>
                                 <img src="assets/chevron.svg" alt="Logo" class="logo" />
                             </button>
                         </li>
+                        <li>
+                            <button onclick="selectDemo('dws-offline-json-asset-alloc-breakdown/demo.html', this)">
+                                [DWS] Highcharts Core + Morningstar Asset Allocation Breakdown (Offline JSON)
```
pawelfus: Rename

demos/index.html
```diff
@@ -240,6 +246,12 @@ <h3>Demos</h3>
                                 <img src="assets/chevron.svg" alt="Logo" class="logo" />
                             </button>
                         </li>
+                        <li>
+                            <button onclick="selectDemo('hypo-offline-json/demo.html', this)">
```
pawelfus: Rename

docs/connectors/morningstar.md
```diff
@@ -94,3 +94,76 @@ Example:
 ## Architecture
 This is a visualization of the Highcharts Morningstar Data Connector:
 ![Highcharts Morningstar Data Connector Architecture](architecture.png)
+
+## Offline mode with `api.json` option
```
pawelfus: Rename

docs/connectors/morningstar.md
```diff
@@ -94,3 +94,76 @@ Example:
...
+
+When `api.json` option is set, it has higher priority than `postman` or online
+API settings.
+
+```js
+const connector = new HighchartsConnectors.Morningstar.SecurityDetailsConnector({
+    api: {
+        json: [{
+            Id: 'OFFLINE_SECURITY',
+            Isin: 'OFFLINE_ISIN',
```
pawelfus: Rename

src/GoalAnalysis/GoalAnalysisConnector.ts
```diff
@@ -105,6 +105,11 @@ export class GoalAnalysisConnector extends MorningstarConnector {
         await super.load();
```
pawelfus: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/RiskScore/RiskScoreConnector.ts
```diff
@@ -254,6 +254,10 @@ export class RiskScoreConnector extends MorningstarConnector {
         const options = this.options;
         const portfolios = options.portfolios;
 
+        if (options.api?.json) {
```
pawelfus, line 254: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/RNANews/RNANewsConnector.ts
```diff
@@ -161,6 +161,11 @@ export class RNANewsConnector extends MorningstarConnector {
         await super.load();
 
         const options = this.options;
+
+        if (options.api?.json) {
```
pawelfus: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/Screeners/InvestmentScreener/InvestmentScreenerConnector.ts
```diff
@@ -87,6 +87,11 @@ export class InvestmentScreenerConnector extends MorningstarConnector {
         await super.load();
 
         const userOptions = { ...this.options, ...options };
+
+        if (userOptions.api?.json) {
```
pawelfus, line 89: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/Screeners/InvestorPreferences/InvestorPreferencesConnector.ts
```diff
@@ -151,6 +151,9 @@ export class InvestorPreferencesConnector extends MorningstarConnector {
         await super.load();
 
         const userOptions = { ...this.options, ...options };
+        if (userOptions.api?.json) {
```
pawelfus, line 153: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/SecurityCompare/SecurityCompareConnector.ts
```diff
@@ -113,6 +113,11 @@ export class SecurityCompareConnector extends MorningstarConnector {
         await super.load();
 
         const userOptions = { ...this.options, ...options };
+
+        if (userOptions.api?.json) {
```
pawelfus, line 114: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/SecurityDetails/SecurityDetailsConnector.ts
```diff
@@ -105,6 +105,11 @@ export class SecurityDetailsConnector extends MorningstarConnector {
         await super.load();
 
         const userOptions = { ...this.options, ...options };
+
+        if (userOptions.api?.json) {
```
pawelfus, line 106: Shouldn't `super.load()` be called after the `options.api?.json` check? Like in TimeSeries connector.

src/Shared/PAUSConnector.ts
```diff
@@ -65,7 +65,7 @@ export abstract class PAUSConnector extends MorningstarConnector {
 
     public override readonly options: PAUSOptions;
 
-    public response?: Response;
+    public responseJSON?: unknown;
```
pawelfus: Q: Do we need this change? Shouldn't be a problem, but if it's public prop, then it should land in the changelog as breaking change.. I mean changing both, the type and the name of the property.

src/TimeSeries/TimeSeriesConnector.ts
```diff
@@ -163,6 +163,9 @@ export class TimeSeriesConnector extends MorningstarConnector {
         await super.load();
 
         const options = this.options;
+        if (options.api?.json) {
```
pawelfus, line 164: As in other connectors: Order `super.load()`

src/XRay/XRayConnector.ts
```diff
@@ -184,6 +184,11 @@ export class XRayConnector extends MorningstarConnector {
         await super.load();
 
         const options = this.options;
+
+        if (options.api?.json) {
```
pawelfus, line 185: As in other connectors: Order `super.load()`

pawelfus (changes requested):
Thanks! Looks really good. Now we can focus on naming

In addition to inline comments, please:
- rename directories

Missing ticket description (any breaking changes? or upgrade notes?)

demos/index.html
```diff
@@ -156,6 +164,8 @@ <h3>Demos</h3>
                             </button>
                         </li>
                         <li>
+                            <button onclick="selectDemo('hypo-pre-fetched-json/demo.html', this)">
```
pawelfus: Something strange with HTML structure here

demos/index.html
```diff
@@ -121,6 +121,14 @@ <h3>Demos</h3>
                             </button>
                         </li>
                         <li>
+                            <button onclick="selectDemo('dws-pre-fetched-json-asset-alloc-breakdown/demo.html', this)">
+                                [DWS] Highcharts Core + Morningstar Asset Allocation Breakdown (Pre-fetched JSON)
+                                <img src="assets/chevron.svg" alt="Logo" class="logo" />
+                            </button>
+                        </li>
+                        <li>
+                            <button onclick="selectDemo('dashboards-rna-news/demo.html', this)" class="selected">
```
pawelfus: Something strange with HTML structure here

tests/Pre-fetched/Pre-FetchedJSON.test.ts
```diff
@@ -0,0 +1,145 @@
+import * as Assert from 'node:assert/strict';
+import '@highcharts/dashboards/es-modules/masters/dashboards.src';
+import * as MC from '../../code/connectors-morningstar.src';
+import * as MCDWS from '../../code/connectors-morningstar-dws.src';
+
+export async function securityDetailsPreFetchedJSONLoad (): Promise<void> {
+    const originalFetch = window.fetch;
```
pawelfus, line 7: Q: Any reason not to do the same in DWS test?

src/Shared/MorningstarConnector.ts
```diff
@@ -79,7 +80,7 @@ export abstract class MorningstarConnector extends External.DataConnector {
         // Expecting async Postman options
         if (!this.api) {
             this.api = new MorningstarAPI(
-                options.postman ?
+                (options.postman) ?
```
pawelfus: Why this change?

src/SecurityDetails/SecurityDetailsConnector.ts
```diff
@@ -135,7 +145,7 @@ export class SecurityDetailsConnector extends MorningstarConnector {
             columns: {},
             id: json[0].Id,
             isin: json[0].Isin,
-            json
+            json: json[0]
```
pawelfus, line 148: Why this change?

src/SecurityCompare/SecurityCompareConnector.ts
```diff
@@ -147,7 +157,7 @@ export class SecurityCompareConnector extends MorningstarConnector {
             columns: {},
             ids: [],
             isins: [],
-            json
+            json: json[0]
```
pawelfus, line 160: Why this change?

src/RNANews/RNANewsConnector.ts
```diff
@@ -203,12 +208,7 @@ export class RNANewsConnector extends MorningstarConnector {
         const response = await api.fetch(url);
         const json = await response.json() as RNANewsJSON.Response;
 
-        this.converter.parse({ json });
-
-        this.getTable().deleteColumns();
-        this.getTable().setColumns(this.converter.getTableColumns());
-
-        return this.applyTableModifiers();
+        return this.parseJSON(json, this.converter);
```
pawelfus: RNAConverted implements it's own `getTableColumns` - and it seems different from all others - could you please double-check this?

pawelfus (changes requested):
Nice! Minor issues to fix

src/Shared/MorningstarConnector.ts
kamil-musialowski: Leftovers, sorry :)

src/SecurityDetails/SecurityDetailsConnector.ts
kamil-musialowski: Previously it wasn't caught by linter - however, I think it is a right change.

SecurityDetails metadata expects `SecurityDetailsJSON.SecurityDetailsResponse` not `Array<SecurityDetailsJSON.SecurityDetailsResponse>`

src/SecurityCompare/SecurityCompareConnector.ts
kamil-musialowski: Same as above

src/RNANews/RNANewsConnector.ts
kamil-musialowski: oohhhh, very good catch!

demos/index.html
```diff
@@ -120,6 +120,12 @@ <h3>Demos</h3>
                                 <img src="assets/chevron.svg" alt="Logo" class="logo" />
                             </button>
                         </li>
+                        <li>
+                            <button onclick="selectDemo('dws-pre-fetched-json-asset-alloc-breakdown/demo.html', this)">
```
pawelfus: Let's add it to closed to the DWS category (after `[DWS] Highcharts Stock + Morningstar Time Series`)

pawelfus (changes requested):
One last thing before we are good to go.

pawelfus (approved):
Thanks!

PR #191, opened 2026-06-12
Adjusted Performance Correlation Matrix converter to new API changes.

jedrzejruta:
Adjusted the Performance Correlation Matrix converter to handle Portfolio entries alongside the updated Morningstar API.

Key points for this change:
- The JSON response now includes Portfolio values, which required adapting converter logic in order to preserve ease of creating heatmap demos along with showing new columns for Portfolio records.
- `includePortfolioCorrelationMatrix` has been renamed to `includePortfolioInCorrelationMatrix`
- The correlation matrix is returned only when the `viewId` is set to `CorrelationMatrix`.
- It is now possible via API to return a `CrossPortfolioCorrelation` (not part of this PR). It is not yet supported in Performance Connector, however shouldn't be hard to add by simply checking for different path (since it shares the same format inside).
- According to OpenAPI, the Benchmark can be skipped completely in the Portfolios array, only valid entries of Holdings are necessary for the payload.
image

---
  - https://app.asana.com/0/0/1215629608352660

### DIFF
--- a/src/Performance/Converters/CorrelationMatrixConverter.ts
+++ b/src/Performance/Converters/CorrelationMatrixConverter.ts
@@ -73,9 +73,7 @@ export class CorrelationMatrixConverter extends MorningstarConverter {
                 for (const key of Correlations) {
                     const { CorrelatedItemKey, SecurityId, Type } = key;
                     const isPortfolio = Type === 'Portfolio';
-                    const name = isPortfolio ?
-                        TrailingTimePeriod + '_Portfolio' + columnSuffix :
-                        TrailingTimePeriod + `_${SecurityId}` + columnSuffix;
+                    const name = `${TrailingTimePeriod}_${isPortfolio ? 'Portfolio' : SecurityId}${columnSuffix}`;
 
                     let correlationIndex = 0;
 
--- a/src/Performance/Converters/CorrelationMatrixConverter.ts
+++ b/src/Performance/Converters/CorrelationMatrixConverter.ts
@@ -68,22 +68,40 @@ export class CorrelationMatrixConverter extends MorningstarConverter {
                 const { TrailingTimePeriod, Correlations } = correlationMatrixItem;
 
                 let rowIndex = 0;
+                let securityIndex = 0;
 
                 for (const key of Correlations) {
-                    const { CorrelatedItemKey, SecurityId, Id } = key;
-                    const name = TrailingTimePeriod + `_${SecurityId}` + columnSuffix;
+                    const { CorrelatedItemKey, SecurityId, Type } = key;
+                    const isPortfolio = Type === 'Portfolio';
+                    const name = isPortfolio ?
+                        TrailingTimePeriod + '_Portfolio' + columnSuffix :
+                        TrailingTimePeriod + `_${SecurityId}` + columnSuffix;
+
+                    let correlationIndex = 0;
 
                     for (let i = 0; i < CorrelatedItemKey.length; i++) {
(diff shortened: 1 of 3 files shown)


pawelfus (approved):
Looks good! Please add Upgrade notes to the PR description (for che changelog).


I compared this PR:
image
and XRayUS:
image

kamil-musialowski (approved):
Good, thanks!

src/Performance/PerformanceOptions.ts
```diff
@@ -58,7 +58,7 @@ export interface PerformanceRequestPayload extends PAUSPayload {
 }
 
 export interface PerformanceRequestSettings extends RequestSettings {
-    includePortfolioCorrelationMatrix?: boolean;
```
PaulDalek, line 61: Question: Shouldn't the deprecation of `includePortfolioCorrelationMatrix` be noted somewhere?

src/Performance/Converters/CorrelationMatrixConverter.ts
```diff
@@ -68,22 +68,40 @@ export class CorrelationMatrixConverter extends MorningstarConverter {
...
 
                 let rowIndex = 0;
+                let securityIndex = 0;
 
                 for (const key of Correlations) {
-                    const { CorrelatedItemKey, SecurityId, Id } = key;
-                    const name = TrailingTimePeriod + `_${SecurityId}` + columnSuffix;
+                    const { CorrelatedItemKey, SecurityId, Type } = key;
+                    const isPortfolio = Type === 'Portfolio';
+                    const name = isPortfolio ?
```
PaulDalek: It can be simplified to:
```
const name = `${TrailingTimePeriod}_${isPortfolio ? 'Portfolio' : SecurityId}${columnSuffix}`;
```

src/Performance/Converters/CorrelationMatrixConverter.ts
```diff
@@ -68,22 +68,40 @@ export class CorrelationMatrixConverter extends MorningstarConverter {
...
+
+                    let correlationIndex = 0;
 
                     for (let i = 0; i < CorrelatedItemKey.length; i++) {
+                        if (CorrelatedItemKey[i].Type === 'Portfolio') {
+                            continue;
+                        }
                         const value = CorrelatedItemKey[i].Value;
-                        table.setCell(name, i, value);
+                        table.setCell(name, correlationIndex, value);
```
PaulDalek, line 85: Question: Should there be `correlationIndex`? It looks like it is always set to 0.

PaulDalek (approved):
Great work! I only left one suggestion and a pair of questions.

src/Performance/Converters/CorrelationMatrixConverter.ts
jedrzejruta: That's not true, the index corresponds to the row index of the column. If i set it explicitly to 0, this is how the data table looks like:
image

src/Performance/PerformanceOptions.ts
jedrzejruta: Good point, I've added a `@deprecated` tag to the old option. Instead, it could use the changelog notes. Which way do you prefer @pawelfus?

DJTechnoo (approved):
LGTM!

src/Performance/PerformanceOptions.ts
pawelfus: I prefer Upgrade notes (in changelog)

PR #192, opened 2026-06-26
enhancement/morningstar-docs-update

PaulDalek:
Expanded main docs page with a better Requirements and Installation sections.

### DIFF
--- a/docs/connectors/morningstar.md
+++ b/docs/connectors/morningstar.md
@@ -13,21 +13,68 @@ API**, with more to come in the future.
 
 ## Requirements
 
-You will need credentials to access the services.
+To use the Morningstar Connectors you need:
+
+- An active [Highcharts licensing](https://www.highcharts.com/license) plus a
+[Morningstar subscription](https://shop.highcharts.com/contact/partner-data).
 
 - Morningstar credentials (this can be either):
   - Access token from your server
   - Username and password
 
-- Morningstar standalone for Highcharts (this can be either):
-  - `@highcharts/connectors-morningstar/connectors-morningstar.js`
-  - `@highcharts/connectors-morningstar/connectors-morningstar-dws.js`
+The connectors ship in two flavours, distinguished by the `dws` suffix. Pick the
+one matching the Morningstar API you intend to call (see [Versions](#versions)):
 
-- Morningstar Connectors for Dashboards (this can be either):
-  - `@highcharts/connectors-morningstar`
-  - `@highcharts/connectors-morningstar/dws`
+- `connectors-morningstar` - standard API.
+- `connectors-morningstar-dws` - newer DWS API.
 
-- Package bundler like Webpack.
+## Installation
+
+You can load the connectors in three equivalent ways. Pick whichever fits your
+project setup. The resulting `HighchartsConnectors.Morningstar.*` and
+`HighchartsConnectors.MorningstarDWS.*` namespaces are the same in all cases.
+
+### 1. ES module (recommended for app projects)
+
+Install the package and import the bundle you need. The import has no default
+export - it registers the connectors as a side effect.
+
+```bash
+npm install @highcharts/connectors-morningstar
+```
+
+```js
+import Highcharts from 'highcharts';
+// Standard API
+import '@highcharts/connectors-morningstar';
+// DWS API
+import '@highcharts/connectors-morningstar/dws';
+```
+
+### 2. UMD bundle through a bundler
+
+If you bundle your app yourself (Webpack, Rollup, esbuild, etc.), point it at
+the UMD build distributed inside the package:
+
+```js
+// Standard API
+@highcharts/connectors-morningstar/connectors-morningstar.js
+// DWS API
+@highcharts/connectors-morningstar/connectors-morningstar-dws.js
+```
+
+### 3. `<script>` tag from the CDN
+
+For quick prototyping or plain HTML pages, include the UMD bundle directly from
+[code.highcharts.com](https://code.highcharts.com):
+
+```html
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<!-- Standard API -->
+<script src="https://code.highcharts.com/connectors/morningstar/connectors-morningstar.js"></script>
+<!-- DWS API -->
+<script src="https://code.highcharts.com/connectors/morningstar/connectors-morningstar-dws.js"></script>
+```
 
 ## Quick Start
 
@@ -36,17 +83,14 @@ products and Highcharts Dashboards.
 
 ### Highcharts Quick Start
 
-You can connect Highcharts core products with Morningstar by using
-`connectors-morningstar.js` or `connectors-morningstar-dws.js` in the
-`connectors-morningstar` bundle. You have to manually create the connector and
-assign the resulting table to your series options.
+After loading a bundle (see [Installation](#installation)), you have to manually
+create the connector and assign the resulting table to your series options.
 
 ### Highcharts Dashboards Quick Start
 
-For Highcharts Dashboards you just need to load the `connectors-morningstar` or
-`connectors-morningstar/dws` bundle, which will register all connectors to the
-Dashboards registry. All Morningstar Connectors are then available in the data
-pool as other connector types.
+After loading a bundle, the Morningstar connectors are registered with the
+Dashboards registry automatically and are then available in the data pool as
+other connector types.
 
 ### Available Connectors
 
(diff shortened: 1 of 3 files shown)


docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
   - Access token from your server
   - Username and password
 
-- Morningstar standalone for Highcharts (this can be either):
-  - `@highcharts/connectors-morningstar/connectors-morningstar.js`
-  - `@highcharts/connectors-morningstar/connectors-morningstar-dws.js`
+The connectors ship in two flavours, distinguished by the `dws` suffix. Pick the
+one matching the Morningstar API you intend to call (see [Versions](#versions)):
+
+- `connectors-morningstar` - standard API.
```
kamil-musialowski: Hmm, I'm thinking maybe we shouldn't call it `standard`?

The DWS is a standard API now, so maybe the old one could be called `legacy API`? 🤔

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
+Install the package and import the bundle you need. The import has no default
+export - it registers the connectors as a side effect.
+
+```bash
+npm install @highcharts/connectors-morningstar
+```
+
+```js
+import Highcharts from 'highcharts';
+// Standard API
```
kamil-musialowski: Change if we decide to drop the `standard` name

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
 
-- Morningstar Connectors for Dashboards (this can be either):
-  - `@highcharts/connectors-morningstar`
-  - `@highcharts/connectors-morningstar/dws`
+If you bundle your app yourself (Webpack, Rollup, esbuild, etc.), point it at
+the UMD build distributed inside the package:
 
-- Package bundler like Webpack.
+```js
+// Standard API
```
kamil-musialowski: and here

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
+```
+
+### 3. `<script>` tag from the CDN
+
+For quick prototyping or plain HTML pages, include the UMD bundle directly from
+[code.highcharts.com](https://code.highcharts.com):
+
+```html
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<!-- Standard API -->
```
kamil-musialowski: and here

docs/connectors/morningstar.md
```diff
@@ -36,17 +83,14 @@ products and Highcharts Dashboards.
 
 ### Highcharts Quick Start
 
-You can connect Highcharts core products with Morningstar by using
-`connectors-morningstar.js` or `connectors-morningstar-dws.js` in the
-`connectors-morningstar` bundle. You have to manually create the connector and
-assign the resulting table to your series options.
+After loading a bundle (see [Installation](#installation)), you have to manually
```
kamil-musialowski, line 86: Maybe add a link pointing to our docs showing an example of simple config and assignment to series data? (or add minimal js example here if that won't be too many lines)

kamil-musialowski (changes requested):
Great job! Only nitpicks and discussions, no errors 👏

jedrzejruta (approved):
Besides Kamil's comments regarding the naming of standard/legacy connectors, nothing to add from my side. Thanks! 👏

docs/connectors/morningstar.md
PaulDalek: Fair point, corrected. Thanks!

docs/connectors/morningstar.md
PaulDalek: Since there is no universal connector that represents all cases (DWS and legacy), I added a short note about the examples from the section below.

kamil-musialowski (approved):
That's better, thank you!

DJTechnoo (approved):
LGTM!

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
 
 ## Requirements
 
-You will need credentials to access the services.
+To use the Morningstar Connectors you need:
+
+- An active [Highcharts licensing](https://www.highcharts.com/license) plus a
+[Morningstar subscription](https://shop.highcharts.com/contact/partner-data).
```
pawelfus: Hm.. Partner Data license includes Highcharts license. Not entirely sure if need that sentence. Is it somehow related to the one of the RAI benchamrks?

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
+The connectors ship in two flavours, distinguished by the `dws` suffix. Pick the
+one matching the Morningstar API you intend to call (see [Versions](#versions)):
+
+- `connectors-morningstar` - legacy API.
+- `connectors-morningstar-dws` - newer DWS API.
+
+## Installation
+
+You can load the connectors in three equivalent ways. Pick whichever fits your
+project setup. The resulting `HighchartsConnectors.Morningstar.*` and
```
pawelfus: It doesn't depend on project setup, but API customer has access to.

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
+`HighchartsConnectors.MorningstarDWS.*` namespaces are the same in all cases.
```
pawelfus: Hm.. not sure what do you mean by "the same". The available connectors are different.

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,68 @@ API**, with more to come in the future.
...
+export - it registers the connectors as a side effect.
+
+```bash
+npm install @highcharts/connectors-morningstar
+```
+
+```js
+import Highcharts from 'highcharts';
+// Legacy API
+import '@highcharts/connectors-morningstar';
```
pawelfus: Just double checking - are you certain it works that way? How then you initialize a connection? Tested with: 

```ts
const connector = new HighchartsConnectors.Morningstar.SecurityDetailsConnector({ ... });
```
And I have error:

Image

pawelfus: React demo uses:
```
import * as MC from '@highcharts/connectors-morningstar';
```

pawelfus (changes requested):
Thanks! Minor changes are needed.

docs/connectors/morningstar.md
PaulDalek: It is basically the same sentence as in the first section, but enriched with URLs. Yes, the changes (the entire Requirements section) are related to the attempt to improve the RAI score, but I think it would be safer to remove this to avoid confusing users. Besides, it is already mentioned earlier. Thanks!

docs/connectors/morningstar.md
PaulDalek: Sure, I corrected the sentence to include the part about API access.

docs/connectors/morningstar.md
PaulDalek: I meant that the `Morningstar` and `MorningstarDWS` namespaces remain the same regardless of how the scripts are loaded. I agree, though, that the original sentence sounded awkward. Corrected.

docs/connectors/morningstar.md
PaulDalek: Ah, right, nice catch! Thanks!

docs/connectors/morningstar.md
```diff
@@ -13,21 +13,67 @@ API**, with more to come in the future.
...
+You can load the connectors in three equivalent ways. Choose whichever best fits
+your project setup. Regardless of the loading method, connectors are exposed
+through the `HighchartsConnectors.Morningstar.*` and
+`HighchartsConnectors.MorningstarDWS.*` namespaces. Whether a connector can
+fetch data depends on your API access.
+
+### 1. ES module (recommended for app projects)
+
+Install the package and import the bundle you need. The import has no default
+export - it registers the connectors as a side effect.
```
pawelfus: Hm.. tricky sentence: there are named exports (as shown below), and it does register as a side effect (when using Highcharts Dashboards). How about something like this?

> Install the package and import the bundle you need. Importing it registers the connectors with Highcharts as a side effect, and also exposes them through the imported namespace (shown below) as named exports.

pawelfus (changes requested):
Almost there, last sentence to fix, see below:

docs/connectors/morningstar.md
PaulDalek: Yes, it makes sense in regard to the previous changes. Thanks, corrected!

pawelfus (approved):
Thanks!

PR #193, opened 2026-07-23
docs/llms-txt-for-crawlers 

AndrzejBuleczka:
Added llms.txt and sitemap.xml with generator script at the docs root for crawlers.


PaulDalek (approved):
Looks fine, thanks! One thing worth to consider is which approach should we take in case of `llm.txt` file:
- One big file for the whole Highcharts documentation.
- A separate file per a documentation section (e.g. Core, Stock, Dash, Connectors, etc.), each referenced in the root `llm.txt`.

AndrzejBuleczka:
Thanks @PaulDalek for check! I'd lean toward option 2 (per-section files), since it matches how we already generate docs per repo/project and keeps each file small and topic-focused, which should help RAI precision. The one caveat: a root file referencing section files isn't part of the `llms.txt` spec, so it only works if our own ingestion follows those references; a plain external crawler just reads `/llms.txt`. So the deciding question is whether this is only for our RAG ingestion or also for public LLM crawlers: if the latter, I'd keep a single spec-compliant root `llms.txt` that sections are merged into.

tools/sitemap.ts
```diff
@@ -0,0 +1,235 @@
...
+
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+
+import { execFileSync } from 'node:child_process';
+import * as FS from 'node:fs';
```
PaulDalek: Import and use specific functions instead:

```
import { existsSync, statSync, writeFileSync } from 'node:fs';
import { join } from 'node:path';
```

tools/sitemap.ts
```diff
@@ -0,0 +1,235 @@
...
+const OUTPUT = Path.join('docs', 'sitemap.xml');
+
+
+/**
+ * Page groups. Each `slug` is appended to BASE_URL for the `<loc>` and, unless
+ * an explicit `file` is given, maps to `docs/connectors/morningstar/<slug>.md`.
+ * `priority` is a soft ranking hint: DWS pages are raised to reflect the
+ * "prefer DWS for Morningstar queries" steer, others default lower.
+ */
+const GROUPS: Array<{
```
PaulDalek: The `GROUPS` array is fine and works, but I’d rather make it dynamic by generating it from the contents of the `docs\connectors` folder. This way, if a new docs page is added, it won’t need to be updated manually.

tools/sitemap.ts
```diff
@@ -0,0 +1,235 @@
...
+ *  Runtime
+ *
+ * */
+
+
+const xml = generate();
+
+FS.writeFileSync(OUTPUT, xml);
+
+const count = (xml.match(/<url>/gu) ?? []).length;
```
PaulDalek: No need for a separate `count`.

PaulDalek (approved):
The second part of the enhancements looks really good! The script nicely creates the `sitemap.xml` file. I’ve left some suggestions below, decide if you want to implement them.

tools/sitemap.ts
```diff
@@ -0,0 +1,235 @@
...
+            { slug: 'dws/equity-style-box' },
+            { slug: 'dws/equity-sectors-breakdown' },
+            { slug: 'dws/equity-residual-risk' },
+            { slug: 'dws/equity-aggregates-residual-risk' },
+            { slug: 'dws/fixed-income-sectors-breakdown' },
+            { slug: 'dws/prospectus-fees' }
+        ]
+    },
+    {
+        title: 'Connectors',
```
pawelfus: Morningstar calls them now `Enterprise Component APIs` (ref: https://developer.morningstar.com/direct-web-services/documentation/enterprise-component-apis/about)

tools/sitemap.ts
```diff
@@ -0,0 +1,235 @@
...
+            { slug: 'hypo-performance' },
+            { slug: 'performance' },
+            { slug: 'risk-score' },
+            { slug: 'security-compare' },
+            { slug: 'security-details' },
+            { slug: 'x-ray' }
+        ]
+    },
+    {
+        title: 'Screeners',
```
pawelfus: Shouldn't this be under EC APIs? @PaulDalek - to reflect the recent structure of docs (eg like this: https://www.highcharts.com/docs/morningstar/screeners/screener)

The same about Time Series below

tools/README.md
```diff
@@ -25,6 +25,14 @@ Release
 
 2. Run `npx ts-node tools/dist --bucket [...] --region [...] --release [x.x.x]`.
 
+Sitemap
+-------
+
+Run `npm run sitemap` (or `npx ts-node tools/sitemap`) to regenerate
+`docs/sitemap.xml`. URLs mirror `docs/llms.txt`; `<lastmod>` is taken per page
+from the last git commit that touched the source markdown, so re-run this after
+editing docs rather than hand-editing the dates.
```
pawelfus: Too easy to forget - can this be added to a post commit hook? That checks for changes in docs, and regenerates when necessary? @PaulDalek  thoughts?

pawelfus (changes requested):
Thanks! A few ideas/improvements

DJTechnoo (approved):
LGTM!

tools/sitemap.ts
```diff
@@ -0,0 +1,235 @@
...
+ *  Generating sitemap.xml for the Morningstar connector docs.
+ *
+ *  URLs mirror docs/llms.txt (base https://www.highcharts.com/docs).
+ *  <lastmod> is derived per page from the last git commit that touched the
+ *  source markdown file, so it stays honest and drives incremental
+ *  re-indexing. Do not hand-write dates - re-run this script instead:
+ *
+ *      npx ts-node tools/sitemap
+ *
+ *  (c) Highsoft AS
```
jedrzejruta, line 9: Generally, I'd rather place it either at the top of the opening comment or right after tool title. 

Also, feel free to add yourself as the author at the bottom of the comment 😉

jedrzejruta (approved):
Great script, works really good!

kamil-musialowski (approved):
Nice! Thanks 👍

tools/sitemap.ts
AndrzejBuleczka: Done, thanks!

tools/sitemap.ts
AndrzejBuleczka: sure thing, refactored, thanks

tools/sitemap.ts
AndrzejBuleczka: true, removed :) thanks

tools/sitemap.ts
AndrzejBuleczka: Thanks, all cleaned up :)

tools/sitemap.ts
AndrzejBuleczka: Thanks for pointing this one :) fixed

tools/sitemap.ts
AndrzejBuleczka: Sounds good to me :)

tools/README.md
AndrzejBuleczka: Good ides, I added husky version

tools/sitemap.ts
PaulDalek: @pawelfus To be honest, in my opinion, it doesn’t really matter, as the links generated in the `sitemap.xml` refer to the correct URLs. Either way, [here](https://github.com/highcharts/connectors-morningstar/pull/193?utm_source=chatgpt.com#discussion_r3815638583), I proposed getting rid of the `GROUPS` completely in favor of reading the docs directly instead.

The only thing I’m wondering is why the pages for [**Return**](https://www.highcharts.com/docs/morningstar/time-series/return), [**Rolling Return**](https://www.highcharts.com/docs/morningstar/time-series/rolling-return), and [**Rating**](https://www.highcharts.com/docs/morningstar/time-series/rating) are not visible in the **Time Series** section of the Highcharts documentation.

tools/libs/DocsPages.ts
```diff
@@ -0,0 +1,403 @@
...
+    '`@highcharts/connectors-morningstar`: the standard API is the package root,',
+    'while the newer DWS API (Investment Details + Time Series) is the `/dws`',
+    'subpath export (`@highcharts/connectors-morningstar/dws`). The UMD/CDN builds',
+    'are shipped as two bundles, `connectors-morningstar.js` and',
+    '`connectors-morningstar-dws.js`. When a query concerns Morningstar, prefer',
+    'the Direct Web Services (DWS) pages below.'
+];
+
+
+export const SECTIONS: DocsSection[] = [
```
PaulDalek: Sorry, but the following solution is just another version of the `GROUPS` approach that I mentioned here: https://github.com/highcharts/connectors-morningstar/pull/193#discussion_r3815638583. Instead, I would propose including the description as a hidden element inside each doc and using it instead of creating an additional object structure, which would have to be updated every time the docs change anyway. What do you think @pawelfus ?

tools/libs/DocsPages.ts
```diff
@@ -0,0 +1,403 @@
+/* *
```
PaulDalek: Also, why placing this file in the `libs` folder?

PaulDalek (changes requested):

tools/libs/DocsPages.ts
AndrzejBuleczka: Sounds good. Happy to read metadata from the docs instead. Which hidden format do you prefer,  comment or --- frontmatter, and will the docs pipeline strip it from the rendered page?
For grouping: OK to just derive it from the folder structure (dws / screeners / time-series / rest), or do you want a specific layout?

tools/sitemap.ts
AndrzejBuleczka: Good catch on Return / Rolling Return / Rating. The question is:  are those missing from the docs sidebar on purpose, or should they be listed?

tools/libs/DocsPages.ts
AndrzejBuleczka: i can move it one level up if you feel it's better

tools/libs/DocsPages.ts
pawelfus: I think better keep it in the documentation files (than maintain a separate structure here). The only thing to add then: when a new page is added, there should be a mechanisms that checks and reports "hey, metainfo for docs is missing".

Folder structure is a good starting point, unless you see a problem there?

tools/sitemap.ts
pawelfus: Should be listed, but that's another repository's problem.

pawelfus (changes requested):
(changing from "comment" to "request changes" so I get notified when changes are ready to review)

tools/libs/DocsPages.ts
PaulDalek: @AndrzejBuleczka When it comes to the doc's metadata, I don't have a specific format in mind, so you can propose something here. As for the pipeline for fetching docs from the repo (if that's what you had in mind), it would be even better to have this information directly in the docs, as it would survive the fetching process. All other information (e.g., header information in the `docs.txt` corpus, such as source, filename, etc.) would be removed. That would had to be added by some sort of post-fetch script but it's not important in this case.

AndrzejBuleczka: ok, I believe i covered both of your requests in the newest refactor, hope it's more solid now

tools/libs/DocsPages.ts
AndrzejBuleczka: after refactor it feels that docs as a setup file actually fits into libs folder, but again, I will adjust to your decision on this one

docs/connectors/morningstar/dws/asset-allocation-breakdown.md
```diff
@@ -1,3 +1,7 @@
+<!-- llms
+description: allocation across asset classes
```
PaulDalek, line 2: I would go through the LLM descriptions once again to make sure they correctly describe the pages.

docs/sitemap.xml
```diff
@@ -0,0 +1,161 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated by tools/sitemap.ts - do not edit by hand. -->
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+
+  <!-- Getting started -->
+  <url>
+    <loc>https://www.highcharts.com/docs/morningstar/morningstar</loc>
+    <lastmod>2026-09-01</lastmod>
+    <priority>0.9</priority>
```
PaulDalek, line 9: I'm not sure about the `priority`, whether it's required as most crawlers and search engines ignore it.

tools/libs/Docs.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/* *
+ *
+ *  Constants
+ *
+ * */
+
+
+export const TITLE = 'Highcharts Connectors for Morningstar';
+
```
PaulDalek, line 80: Why is there so much space between the `consts`?

tools/libs/Docs.ts
```diff
@@ -0,0 +1,290 @@
...
+        description: 'example integrations'
+    }
+];
+
+
+/**
+ * Section layout. Headings and priorities live here; the pages are read from
+ * the folders, so adding a doc never means touching this list.
+ */
+const LAYOUT: Layout[] = [
```
PaulDalek, line 125: I'm still not convinced about having this static layout, as the `priority` can be dropped, `dir` and `groups` can be taken from the folder structure and the `heading` can be stored along with the hidden LLM comments. However, the scripts work correctly, so in my opinion, there’s no need to change that. Up to you.

tools/libs/Docs.ts
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * URL slug for a source file relative to `DOCS_ROOT`.
+ *
+ * @param rel
+ * Path relative to `DOCS_ROOT`.
+ *
+ * @return
+ * URL slug.
+ */
+function slugFor (
```
PaulDalek, line 156: No need to have this as a separate utility function.

tools/llms.ts
```diff
@@ -0,0 +1,80 @@
...
+
+
+/* *
+ *
+ *  Functions
+ *
+ * */
+
+
+function item (
```
PaulDalek, line 44: I would add some JSDoc comment to this function.

tools/llms.ts
```diff
@@ -0,0 +1,80 @@
...
+    page: DocsPage
+): string {
+    return `- [${page.title}](${page.url}): ${page.description}`;
+}
+
+
+function generate (): string {
```
PaulDalek, line 51: I would add some JSDoc comment to this function too.

tools/sitemap.ts
```diff
@@ -0,0 +1,157 @@
...
+            `    <loc>${page.url}</loc>`,
+            `    <lastmod>${lastModified(page.file)}</lastmod>`,
+            `    <priority>${section.priority}</priority>`,
+            '  </url>'
+        ])
+    ];
+}
+
+
+function generate (): string {
```
PaulDalek, line 131: Finally, I would add some JSDoc comment to this function as well.

tools/llms.ts
```diff
@@ -0,0 +1,80 @@
...
+
+
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+
+import { writeFileSync } from 'node:fs';
```
PaulDalek, line 20: This is a matter of preference but both scripts (`llms.ts` and `sitemap.ts`) could be combined into one, as they are quite small. On the other hand, keeping them separate provides some clarity. Up to you.

tools/libs/Docs.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+
+/**
+ * Recursively lists every markdown file under `DOCS_ROOT`.
+ *
+ * @return
+ * Paths relative to `DOCS_ROOT`.
+ */
+function allDocs (): string[] {
```
PaulDalek, line 234: No need to have this as a separate utility function.

PaulDalek (approved):
Looks much better, thanks! I found a few more things but I approve it to not block this PR.

As for the hidden LLM descriptions on each documentation page, I think they can stay as they are for now. It's better than having a fixed structure object, but bear in mind that this might change in the future, just like any other additional elements added for RAI or MCP usage.

docs/sitemap.xml
AndrzejBuleczka: good point, removed

tools/libs/Docs.ts
AndrzejBuleczka: agreed, fixed

tools/llms.ts
AndrzejBuleczka: doclet added

tools/llms.ts
AndrzejBuleczka: doclet added

tools/sitemap.ts
AndrzejBuleczka: doclet added

tools/llms.ts
AndrzejBuleczka: I prefer to have them separate

tools/libs/Docs.ts
AndrzejBuleczka: granted, refactored and fixed

tools/libs/Docs.ts
AndrzejBuleczka: not needed, fixed

docs/connectors/morningstar/dws/asset-allocation-breakdown.md
AndrzejBuleczka: good intuition, I improved some of them

tools/libs/Docs.ts
AndrzejBuleczka: I dropped the `priority`, you're right that crawlers ignore it, and it slimmed the layout down too.

I kept `heading`/`dir`/`groups` though, because the folder structure doesn't carry that info:

- On disk `dws`, `screeners` and `time-series` are all siblings, but in the output Screeners and Time Series are nested **under** Enterprise Components while DWS is top-level — that nesting is an editorial choice.
- "Enterprise Components" has no folder at all; it's the direct `.md` children of `morningstar/`.

So all that's left is the section names, their order and nesting. The pages still come entirely from the folders, so adding a doc never means touching this list.

AndrzejBuleczka:
Hi @pawelfus , do you want to check it on your own? All of @PaulDalek comments are now addressed.

pawelfus (commented):
On hold for now, testing the json-ld approach instead

PR #194, opened 2026-08-05
[DO NOT MERGE] Added 'create your first chart' article.

kamil-musialowski:
It should be easier to review it here, rather than Google Docs I guess.

### DIFF
--- a/docs/connectors/creating-your-first-chart.md
+++ b/docs/connectors/creating-your-first-chart.md
@@ -1,0 +1,315 @@
+# Creating Your First Chart from Morningstar Market Data
+
+Morningstar is one of the most trusted names in financial data. But getting that data onto a screen has traditionally meant handling complex requests, and a mapping layer that someone has to maintain forever.
+
+The Highcharts connectors for Morningstar Direct Web Services (DWS) remove that detour. You configure your credentials once, pick the data you want, and the connector hands you a formatted data table that Highcharts can render directly, with easy access to raw data as well.
+
+In this tutorial we will go from an empty HTML file to a working chart built on Morningstar data. We will use the DWS bundle, which targets Morningstar's newer API, and we will finish by looking at how the same few lines of setup unlock the rest of the DWS data catalogue.
+
+## What we will build
+
+An asset allocation chart showing how a fund splits its net assets across equity, bonds, cash, and other holdings — built with the **Investment Details Connector** and Highcharts, in four steps from an empty file.
+
+The connectors ship in two flavours, and they are not interchangeable:
+
+- **`connectors-morningstar`** — the legacy API, covering Security Details, Screeners, X-Ray, Risk Score, Performance, and more.
+- **`connectors-morningstar-dws`** — the newer DWS API, covering the **Investment Details API** and the **Time Series API**, with more to come.
+
+This tutorial uses the DWS bundle throughout. Its connectors live under the `HighchartsConnectors.MorningstarDWS` namespace, while the legacy ones live under `HighchartsConnectors.Morningstar`.
+
+DWS exposes two connectors: the Investment Details Connector we are about to use, which covers portfolio composition, and the Time Series Connector for historical data such as performance and fees.
+
+
+## Building the asset allocation chart
+
+### Step 1: Load Highcharts and the DWS connectors
+
+Create an `index.html` file. Asset allocation is a snapshot rather than a time series, so plain Highcharts is all we need here:
+
+```html
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8" />
+    <title>My First Morningstar Chart</title>
+    <script src="https://code.highcharts.com/highcharts.js"></script>
+    <script src="https://code.highcharts.com/connectors/morningstar/connectors-morningstar-dws.js"></script>
+</head>
+<body>
+    <div id="container"></div>
+    <script src="./demo.js"></script>
+</body>
+</html>
+```
+
+Loading the bundle registers the connectors with Highcharts as a side effect and exposes them on the global `HighchartsConnectors` object. That is all the wiring there is.
+
+If you are working in an app project rather than a plain HTML page, install the package and import it instead:
+
+```bash
+npm install @highcharts/connectors-morningstar
+```
+
+```js
+import Highcharts from 'highcharts';
+import { InvestmentsConnector } from '@highcharts/connectors-morningstar/dws';
+```
+
+One difference to keep in mind as you read on. The `HighchartsConnectors.MorningstarDWS` namespace used throughout this tutorial comes from the UMD bundle; the ES module exports its connectors directly, so in an app project you write `new InvestmentsConnector({ ... })` rather than going through a global.
+
+### Step 2: Configure the Investment Details Connector
+
+Now create `demo.js`. Because loading data is asynchronous, we will build the chart inside an `async` function and call it at the end of the file. The `InvestmentsConnector` describes a single security, and you choose what you want to know about it by naming one or more **converters**:
+
+```js
+async function createChart () {
+    const connector = new HighchartsConnectors.MorningstarDWS.InvestmentsConnector({
+        api: {
+            access: {
+                token: 'your_access_token'
+            }
+        },
+        security: {
+            id: '0P00000FIA'
+        },
+        converters: {
+            AssetAllocationBreakdown: {}
+        }
+    });
+    ...
+```
+
+Three of those options deserve a closer look.
+
+**`security`** takes a single `id` — the Investment Details API describes one investment at a time. The identifier above is a Morningstar `performanceId`.
+
+**`converters`** is the heart of this connector. Rather than one connector per data type, a single instance can request several converters at once, and each one populates its own named data tables. Here we ask for just `AssetAllocationBreakdown`; the empty object is not a placeholder to fill in, since most converters need no options at all. Some accept `startDate` and `endDate` to bound a historical series:
+
+```js
+converters: {
+    AssetAllocationBreakdown: {},
+    EquitySectorsBreakdown: {},
+    EquityStyleBox: {
+        startDate: '2025-01-01',
+        endDate: '2025-12-01'
+    }
+}
+```
+
+That is one request block and one `load()` for three datasets — a fund profile page with six panels needs one connector with six converters, not six connectors.
+
+**`api.access`** carries your credentials. If you prefer a username and password over a token, note that DWS uses its own dedicated fields:
+
+```js
+api: {
+    access: {
+        dwsUsername: 'your_username',
+        dwsPassword: 'your_password'
+    }
+}
+```
+### Step 3: Load and read the data tables and render
+
+This is the step where the connector's design pays off. `AssetAllocationBreakdown` populates three data tables, already shaped for Highcharts, each retrieved by name:
+
+```js
+    await connector.load();
+```
+
+```js
+    const generalTable = connector.getTable('AssetAlloc'),
+        canadaTable = connector.getTable('CanadianAssetAlloc'),
+        underlyingTable = connector.getTable('UnderlyingAssetAlloc');
+```
+
+`AssetAlloc` is the general view, splitting net assets across bonds, cash, convertible bonds, equity, and other. `CanadianAssetAlloc` is the Canadian view, and `UnderlyingAssetAlloc` breaks the portfolio down by underlying instrument type.
+
+Every table has a `Type` column naming its categories. `AssetAlloc` and `CanadianAssetAlloc` then carry one column per measure — `Long`, `LongRescaled`, `Net`, and `Short` — while `UnderlyingAssetAlloc` holds a single `UnderlyingInstruments` column of percentages. Because this is snapshot data rather than a series, you hand whole columns to Highcharts: `Type` becomes the axis categories, and each measure becomes a series:
+
+```js
+    ...
+    Highcharts.chart('container', {
+        chart: {
+            type: 'column'
+        },
+        title: {
+            text: 'General Asset Allocation Breakdown'
+        },
+        yAxis: {
+            labels: {
+                format: '{value}%'
+            }
+        },
+        xAxis: {
+            categories: generalTable.getColumn('Type')
+        },
+        series: [{
+            name: 'Long',
+            data: generalTable.getColumn('Long')
+        }, {
+            name: 'Long Rescaled',
+            data: generalTable.getColumn('LongRescaled')
+        }, {
+            name: 'Net',
+            data: generalTable.getColumn('Net')
+        }, {
+            name: 'Short',
+            data: generalTable.getColumn('Short')
+        }]
+    });
+}
+
+createChart();
+```
+
+That is your first chart from Morningstar data: one connector, one converter, and no mapping layer in between.
+
+Two more details worth knowing. `AssetAlloc` also carries US and non-US splits under prefixed column names — `Us_Long`, `NonUs_Net`, and so on — if you want to break the allocation down geographically without a second request. And the connector exposes the security's `performanceId` through its metadata, which is handy for a subtitle:
+
+```js
+subtitle: {
+    text: `Performance ID: ${connector.metadata.AssetAllocationBreakdown.performanceId}`
+}
+```
+
+## The other connector: Time Series
+
+Portfolio composition is one half of DWS. The other is the `TimeSeriesConnector`, which targets Morningstar's Time Series API and returns historical data for up to 25 securities in a single request.
+
+It follows the same three beats you just used — and, like the code above, it belongs inside an `async` function. Two things differ: instead of converters you pick data with a `category` and a `dataPoint`, and instead of one security you pass a list:
+
+```js
+const growthConnector = new HighchartsConnectors.MorningstarDWS.TimeSeriesConnector({
+    api: {
+        access: {
+            token: 'your_access_token'
+        }
+    },
+    ids: [{
+        id: '0P00000FIA',
+        idType: 'performanceId'
+    }, {
+        id: '0P00002PB8',
+        idType: 'performanceId'
+    }],
+    category: 'performance',
+    dataPoint: 'growth',
+    startDate: '2024-10-30',
+    endDate: '2025-10-30',
+    currencyId: 'EUR'
+});
+
+await growthConnector.load();
+```
+
+`category` and `dataPoint` map directly onto Morningstar's Time Series API paths: take the two segments immediately after `time-series/v1/`, so `.../time-series/v1/performance/growth/` becomes `category: 'performance'` and `dataPoint: 'growth'`. The available categories are `performance`, `fees-expenses`, `portfolio-holdings`, `portfolio-analytics`, `fund-research`, `fund-sustainability`, `corporate-actions`, and `reference`. `performanceId` is the default `idType`, so you can omit it, and `currencyId` takes an ISO 4217 code — worth setting explicitly when you compare funds domiciled in different markets, since it otherwise defaults to each investment's own base currency.
+
+Reading the result differs from the tables above. There is a single table whose first column is `Date`, holding `yyyy-MM-dd` strings sorted ascending. Request one security and the values land in a `Value` column; request several and each is suffixed with that security's `performanceId` — always the `performanceId`, whatever `idType` you asked with. Pair `Date` with a value column to get `[date, value]` rows, and Highcharts Stock takes them directly:
+
+```js
+Highcharts.stockChart('container', {
+    series: [{
+        name: 'Capital Group Global Equity Fund (LUX) B',
+        data: growthConnector.getTable().getRows(
+            void 0,
+            void 0,
+            ['Date', 'Value_0P00000FIA']
+        )
+    }]
+});
+```
+
+`getRows()` takes a row offset and a row count, both skipped with `void 0` here, then the columns you want in the order you want them. Note the different bundle: a time series wants `highstock.js` rather than the `highcharts.js` we loaded in Step 1.
+
+## What else DWS gives you
+
+`AssetAllocationBreakdown` is one of eight converters currently available on the Investment Details Connector. The rest follow the identical pattern — name the converter, read its tables:
+
+- **Asset Allocation Breakdown** — allocation across asset classes, with general, Canadian, and underlying-instrument views.
+- **Country and Regional Exposure Breakdown** — geographic exposure by region and country, for equity, fixed income, and revenue.
+- **Equity Sectors Breakdown** — sector exposure at super-sector, sector, and industry levels, in the `EqSuperSectors`, `EqSectors`, and `EqIndustries` tables.
+- **Fixed Income Sectors Breakdown** — fixed-income sector exposure across super, primary, and secondary sectors, plus per-region tables.
+- **Equity Style Box** — Morningstar's proprietary style box and stock grades, as both a current grid and a historical series.
+- **Equity Residual Risk and Return Sensitivity** — Alpha, Beta, and RSquare at daily and monthly granularity.
+- **Equity Aggregates Residual Risk and Return Sensitivity** — aggregate residual-risk statistics with their company counts.
+- **Prospectus Fees** — management, administration, and distribution fees, sales loads, and expense ratios.
+
+
+## Two things that will save you time
+
+**Develop against pre-fetched JSON.** Both DWS connectors accept an `api.json` option that bypasses authentication and the network entirely, using a payload you supply. It takes priority over `postman` and online API settings, which makes it ideal for local development, demos, and tests where you do not want to burn API calls:
+
+```js
+const connector = new HighchartsConnectors.MorningstarDWS.InvestmentsConnector({
+    api: {
+        json: {
+            AssetAllocationBreakdown: {
+                assetAllocationBreakdown: {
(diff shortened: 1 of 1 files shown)


DJTechnoo (approved):
Beautifully written! 🖊️📖🪶

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,315 @@
...
+This tutorial uses the DWS bundle throughout. Its connectors live under the `HighchartsConnectors.MorningstarDWS` namespace, while the legacy ones live under `HighchartsConnectors.Morningstar`.
+
+DWS exposes two connectors: the Investment Details Connector we are about to use, which covers portfolio composition, and the Time Series Connector for historical data such as performance and fees.
+
+
+## Building the asset allocation chart
+
+### Step 1: Load Highcharts and the DWS connectors
+
+Create an `index.html` file. Asset allocation is a snapshot rather than a time series, so plain Highcharts is all we need here:
```
jedrzejruta: Nitpicking, but I'd use other wording for the phrase `plain Highcharts` or add a short explanation for what a "plain Highcharts" is. What do you think?

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,315 @@
...
+```bash
+npm install @highcharts/connectors-morningstar
+```
+
+```js
+import Highcharts from 'highcharts';
+import { InvestmentsConnector } from '@highcharts/connectors-morningstar/dws';
+```
+
+One difference to keep in mind as you read on. The `HighchartsConnectors.MorningstarDWS` namespace used throughout this tutorial comes from the UMD bundle; the ES module exports its connectors directly, so in an app project you write `new InvestmentsConnector({ ... })` rather than going through a global.
```
jedrzejruta: Not sure if everyone reading would understand phrase "through a global". Maybe it's worth adding a further explanation.

Also, we could use some backlinking to the docs article explaining bundles in Highcharts.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Step 2: Configure the Investment Details Connector
+
+Now create `demo.js`. Because loading data is asynchronous, we will build the chart inside an `async` function and call it at the end of the file. The `InvestmentsConnector` describes a single security, and you choose what you want to know about it by naming one or more **converters**:
```
jedrzejruta:
```suggestion
Now create `demo.js`. Because loading data is asynchronous, we will build the chart inside an `async` function and call it at the end of the file. The `InvestmentsConnector` describes a single security, and you choose what you want to know about it by naming one or more of available **converters**:
```

What do you think?

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,315 @@
...
+    await connector.load();
+```
+
+```js
+    const generalTable = connector.getTable('AssetAlloc'),
+        canadaTable = connector.getTable('CanadianAssetAlloc'),
+        underlyingTable = connector.getTable('UnderlyingAssetAlloc');
+```
+
+`AssetAlloc` is the general view, splitting net assets across bonds, cash, convertible bonds, equity, and other. `CanadianAssetAlloc` is the Canadian view, and `UnderlyingAssetAlloc` breaks the portfolio down by underlying instrument type.
```
jedrzejruta:
```suggestion
`AssetAlloc` is the general view, splitting net assets across bonds, cash, convertible bonds, equity and other. `CanadianAssetAlloc` is the Canadian view, and `UnderlyingAssetAlloc` breaks the portfolio down by underlying instrument type.
```

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,315 @@
...
+        data: growthConnector.getTable().getRows(
+            void 0,
+            void 0,
+            ['Date', 'Value_0P00000FIA']
+        )
+    }]
+});
+```
+
+`getRows()` takes a row offset and a row count, both skipped with `void 0` here, then the columns you want in the order you want them. Note the different bundle: a time series wants `highstock.js` rather than the `highcharts.js` we loaded in Step 1.
```
jedrzejruta: I know it's yet not fully supported in Connectors, but for the future it would be nice to add a showcase of latest Highcharts DataTable support.

jedrzejruta (changes requested):
Fantastic writing! 🎉 

Couple of nitpicks/ideas for the article below:

docs/connectors/creating-your-first-chart.md
kamil-musialowski: Hmm, you're right - from users perspective it is confusing. Should be `Highcharts Core`.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+
+Morningstar is one of the most trusted names in financial data. But getting that data onto a screen has traditionally meant handling complex requests, and a mapping layer that someone has to maintain forever.
+
+The Highcharts connectors for Morningstar Direct Web Services (DWS) remove that detour. You configure your credentials once, pick the data you want, and the connector hands you a formatted data table that Highcharts can render directly, with easy access to raw data as well.
+
+In this tutorial we will go from an empty HTML file to a working chart built on Morningstar data. We will use the DWS bundle, which targets Morningstar's newer API, and we will finish by looking at how the same few lines of setup unlock the rest of the DWS data catalogue.
+
+## What we will build
+
+An asset allocation chart showing how a fund splits its net assets across equity, bonds, cash, and other holdings — built with the **Investment Details Connector** and Highcharts, in four steps from an empty file.
```
PaulDalek: I would replace all occurrences of the em dash (`—`) with the hyphen (`-`). The former is overused by AI models :).

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+## Building the asset allocation chart
+
+### Step 1: Load Highcharts and the DWS connectors
+
+Create an `index.html` file. Asset allocation is a snapshot rather than a time series, so Highcharts Core is all we need here:
+
+```html
+<!DOCTYPE html>
+<html>
+<head>
```
PaulDalek: This is a matter of preference, but I would add one extra level of indentation for the content inside the `<html>` tag.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+        dwsPassword: 'your_password'
+    }
+}
+```
+### Step 3: Load and read the data tables and render
+
+This is the step where the connector's design pays off. `AssetAllocationBreakdown` populates three data tables, already shaped for Highcharts, each retrieved by name:
+
+```js
+    await connector.load();
```
PaulDalek: Here, on the other hand, I would remove the indentation.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+```js
+    const generalTable = connector.getTable('AssetAlloc'),
```
PaulDalek: Here too ;)

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+            data: generalTable.getColumn('LongRescaled')
+        }, {
+            name: 'Net',
+            data: generalTable.getColumn('Net')
+        }, {
+            name: 'Short',
+            data: generalTable.getColumn('Short')
+        }]
+    });
+}
```
PaulDalek, line 164: Is this closing curly bracket intentional? If not, I would remove it and the indentation from the entire code block as well.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+The connectors ship in two flavours, and they are not interchangeable:
+
+- **`connectors-morningstar`** — the legacy API, covering Security Details, Screeners, X-Ray, Risk Score, Performance, and more.
+- **`connectors-morningstar-dws`** — the newer DWS API, covering the **Investment Details API** and the **Time Series API**, with more to come.
+
+This tutorial uses the DWS bundle throughout. Its connectors live under the `HighchartsConnectors.MorningstarDWS` namespace, while the legacy ones live under `HighchartsConnectors.Morningstar`.
+
+DWS exposes two connectors: the Investment Details Connector we are about to use, which covers portfolio composition, and the Time Series Connector for historical data such as performance and fees.
+
+
```
PaulDalek, line 21: No need for the extra blank line.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+```js
+api: {
+    access: {
+        dwsUsername: 'your_username',
+        dwsPassword: 'your_password'
+    }
+}
+```
+### Step 3: Load and read the data tables and render
```
PaulDalek: I would add a blank line before the `### Step 3` section.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,317 @@
...
+- **Asset Allocation Breakdown** — allocation across asset classes, with general, Canadian, and underlying-instrument views.
+- **Country and Regional Exposure Breakdown** — geographic exposure by region and country, for equity, fixed income, and revenue.
+- **Equity Sectors Breakdown** — sector exposure at super-sector, sector, and industry levels, in the `EqSuperSectors`, `EqSectors`, and `EqIndustries` tables.
+- **Fixed Income Sectors Breakdown** — fixed-income sector exposure across super, primary, and secondary sectors, plus per-region tables.
+- **Equity Style Box** — Morningstar's proprietary style box and stock grades, as both a current grid and a historical series.
+- **Equity Residual Risk and Return Sensitivity** — Alpha, Beta, and RSquare at daily and monthly granularity.
+- **Equity Aggregates Residual Risk and Return Sensitivity** — aggregate residual-risk statistics with their company counts.
+- **Prospectus Fees** — management, administration, and distribution fees, sales loads, and expense ratios.
+
+
```
PaulDalek, line 243: No need for the extra blank line here too.

docs/connectors/creating-your-first-chart.md
kamil-musialowski: Yes, it's a followup to curly bracket opening at 64, since code is splitted into two parts. Although I'm not sure how readable that is.

PaulDalek (approved):
Great work, thanks!

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+```js
+import Highcharts from 'highcharts';
+import { InvestmentsConnector } from '@highcharts/connectors-morningstar/dws';
+```
+
+One difference to keep in mind as you read on. The `<script>` tag exposes everything on the global `HighchartsConnectors` object, which is why this tutorial writes `new HighchartsConnectors.MorningstarDWS.InvestmentsConnector({ ... })`. An ES module sets no such global, so there you write `new InvestmentsConnector({ ... })` with whatever you imported. Everything else is the same. See [Installation](https://www.highcharts.com/docs/getting-started/installation) for more on the two loading methods.
+
+### Step 2: Configure the Investment Details Connector
+
+Now create `demo.js`. Because loading data is asynchronous, we will build the chart inside an `async` function and call it at the end of the file. The `InvestmentsConnector` describes a single security, and you choose what you want to know about it by naming one or more of the available **converters**:
```
pawelfus: `converters` -> it could be a nice place for a link to other available converters in this connector.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+        converters: {
+            AssetAllocationBreakdown: {}
+        }
+    });
+    ...
+```
+
+Three of those options deserve a closer look.
+
+**`security`** takes a single `id` - the Investment Details API describes one investment at a time. The identifier above is a Morningstar `performanceId`.
```
pawelfus, line 83: Q: Should we mention other IDs that can be used? ISIN, CUSIP etc?

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+**`api.access`** carries your credentials. If you prefer a username and password over a token, note that DWS uses its own dedicated fields:
+
+```js
+api: {
+    access: {
+        dwsUsername: 'your_username',
+        dwsPassword: 'your_password'
+    }
+}
+```
```
pawelfus, line 111: We should not advise that way of fetching the data, please remove

pawelfus: And add details how to obtain the token

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+
+That is your first chart from Morningstar data: one connector, one converter, and no mapping layer in between.
+
+Two more details worth knowing. `AssetAlloc` also carries US and non-US splits under prefixed column names - `Us_Long`, `NonUs_Net`, and so on - if you want to break the allocation down geographically without a second request. And the connector exposes the security's `performanceId` through its metadata, which is handy for a subtitle:
+
+```js
+subtitle: {
+    text: `Performance ID: ${connector.metadata.AssetAllocationBreakdown.performanceId}`
+}
+```
```
pawelfus, line 180: Q: Should we add a paragraph with screenshot of the expected chart? Or, a jsfiddle/codepen demo with code in action (using our proxy server)

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
+# Creating Your First Chart from Morningstar Market Data
```
pawelfus: The title suggest Market data, meanwhile article is about Funds data

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+
+```js
+subtitle: {
+    text: `Performance ID: ${connector.metadata.AssetAllocationBreakdown.performanceId}`
+}
+```
+
+## The other connector: Time Series
+
+Portfolio composition is one half of DWS. The other is the `TimeSeriesConnector`, which targets Morningstar's Time Series API and returns historical data for up to 25 securities in a single request.
```
pawelfus: > Portfolio composition is one half of DWS

Nowadays, there is also Scenario analysis, Screeners and Portfolio-related tools. In the first part of the article we discuss not exactly the *portfolio* composition, but investment composition. Not sure if that's a big difference, but perhaps we should use wording from Morningstar docs.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+    dataPoint: 'growth',
+    startDate: '2024-10-30',
+    endDate: '2025-10-30',
+    currencyId: 'EUR'
+});
+
+await growthConnector.load();
+```
+
+`category` and `dataPoint` map directly onto Morningstar's Time Series API paths: take the two segments immediately after `time-series/v1/`, so `.../time-series/v1/performance/growth/` becomes `category: 'performance'` and `dataPoint: 'growth'`. The available categories are `performance`, `fees-expenses`, `portfolio-holdings`, `portfolio-analytics`, `fund-research`, `fund-sustainability`, `corporate-actions`, and `reference`. `performanceId` is the default `idType`, so you can omit it, and `currencyId` takes an ISO 4217 code - worth setting explicitly when you compare funds domiciled in different markets, since it otherwise defaults to each investment's own base currency.
```
pawelfus: "he available categories are" -> Listing categories: better limit to a few and add "and more." New categories will be (probably) added in the future.

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+
+Reading the result differs from the tables above. There is a single table whose first column is `Date`, holding `yyyy-MM-dd` strings sorted ascending. Request one security and the values land in a `Value` column; request several and each is suffixed with that security's `performanceId` - always the `performanceId`, whatever `idType` you asked with. Pair `Date` with a value column to get `[date, value]` rows, and Highcharts Stock takes them directly:
+
+```js
+Highcharts.stockChart('container', {
+    series: [{
+        name: 'Capital Group Global Equity Fund (LUX) B',
+        data: growthConnector.getTable().getRows(
```
pawelfus: I think we should consider `dataTable` + `dataMapping` option here. What do you think @jedrzejruta - do you see any blockers?

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+    underlyingTable = connector.getTable('UnderlyingAssetAlloc');
+```
+
+`AssetAlloc` is the general view, splitting net assets across bonds, cash, convertible bonds, equity, and other. `CanadianAssetAlloc` is the Canadian view, and `UnderlyingAssetAlloc` breaks the portfolio down by underlying instrument type.
+
+Every table has a `Type` column naming its categories. `AssetAlloc` and `CanadianAssetAlloc` then carry one column per measure - `Long`, `LongRescaled`, `Net`, and `Short` - while `UnderlyingAssetAlloc` holds a single `UnderlyingInstruments` column of percentages. Because this is snapshot data rather than a series, you hand whole columns to Highcharts: `Type` becomes the axis categories, and each measure becomes a series:
+
+```js
+    ...
+    Highcharts.chart('container', {
```
pawelfus, line 135: I think we should consider `dataTable` + `dataMapping` (on a series level) option here. What do you think @jedrzejruta - do you see any blockers?

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+});
+```
+
+`getRows()` takes a row offset and a row count, both skipped with `void 0` here, then the columns you want in the order you want them. Note the different bundle: a time series wants `highstock.js` rather than the `highcharts.js` we loaded in Step 1.
+
+Highcharts 13 added a chart-level [`dataTable`](https://api.highcharts.com/highcharts/dataTable) option that accepts a `DataTable` instance directly, with [`series.dataMapping`](https://api.highcharts.com/highcharts/plotOptions.series.dataMapping) binding its columns to point properties - so a connector's table could be handed to the chart whole, without picking it apart first. The connectors target Highcharts 12 for now, so that route is not available yet.
+
+## What else DWS gives you
+
+`AssetAllocationBreakdown` is one of eight converters currently available on the Investment Details Connector. The rest follow the identical pattern - name the converter, read its tables:
```
pawelfus: Hm.. I'm a bit confused. Why now going back to `AssetAllocation` demo? Should we consider removing the whole `The other connector: Time Series` part? Does not quite fit the whole article now. Or perhaps move the paragraph closer to the converters discussion? Or just remove, and keep [link](https://www.highcharts.com/docs/morningstar/dws/investments-details-connector#available-data-converters) to avail converters in one of the first paragraphs (already mentioned)

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+            converters: {
+                ProspectusFees: {}
+            }
+        }]
+    }
+    // ...gui and components
+});
+```
+
+Components then bind to a table by name through `dataTableKey`, so a Grid or KPI reads the connector's output the same way the chart above did. `ProspectusFees` returns a single row, so the documentation pairs it with the `Invert` data modifier to turn it into a label/value list - worth knowing when a table's shape does not match the component you have in mind.
```
pawelfus: Perhaps `ProspectusFees` is not the best choice for the first-chart article? How about one of the other breakdowns (country, regional or sectors)?

pawelfus (changes requested):
Really good article! A few comments/questions. I'm open for a discussion

docs/connectors/creating-your-first-chart.md
```diff
@@ -0,0 +1,316 @@
...
+
+In this tutorial we will go from an empty HTML file to a working chart built on Morningstar data. We will use the DWS bundle, which targets Morningstar's newer API, and we will finish by looking at how the same few lines of setup unlock the rest of the DWS data catalogue.
+
+## What we will build
+
+An asset allocation chart showing how a fund splits its net assets across equity, bonds, cash, and other holdings - built with the **Investment Details Connector** and Highcharts, in three steps from an empty file.
+
+The connectors ship in two flavours, and they are not interchangeable:
+
+- **`connectors-morningstar`** - the legacy API, covering Security Details, Screeners, X-Ray, Risk Score, Performance, and more.
```
PaulDalek: Both the Morningstar and Highcharts docs refer to the legacy API as **Enterprise Component APIs** now, so I think it would be worth noting that here.

PaulDalek (approved):
Other than what was mentioned, everything looks correct! Only one minor thing.

docs/connectors/creating-your-first-chart.md
kamil-musialowski: Yes we should BUT...

...we don't support other ID types in InvestmentsConnector 😬 Should that be reported as a feature request in this repo and only then added to the article?

docs/connectors/creating-your-first-chart.md
kamil-musialowski: I've added it because firstly, it has a bit different data structure than others, secondly it's the only other DWSConnector that we have. I've rearranged the paragraphs order, so you'll see if that's better or it should be completely removed.

docs/connectors/creating-your-first-chart.md
pawelfus: Ah, right! Yes, please report it. We will keep the article as is, and will not wait for this feature. It is supported in TimeSeries, tested with:

```
    const growthConnector =
        new HighchartsConnectors.MorningstarDWS.TimeSeriesConnector({
            api: {
                access: {
                    token: token.access_token
                }
            },
            ids: [{
                id: 'LU2607536815',
                idType: 'isin'
            }],
            category: 'performance',
            dataPoint: 'growth',
            startDate: '2024-10-30',
            endDate: '2025-10-30',
            currencyId: 'EUR'
        });

    await growthConnector.load();
    console.log('Growth data', growthConnector);
 ```

Comment (kamil-musialowski) [reply docs/connectors/creating-your-first-chart.md:83]: Yes, I've seen that TimeSeries is supported - but thanks for double checking that. Reported here https://github.com/highcharts/connectors-morningstar/issues/200

File: docs/connectors/creating-your-first-chart.md (hunk 21 above)
Comment (kamil-musialowski) [reply docs/connectors/creating-your-first-chart.md]: If I remember correctly, dataTable + dataMapping requires HC 13, while connectors require HC ^12.6.0.

Of course, I can add it to article but I don't want to confuse the users. So I'd suggest either a hint, that dataTable approach is available from v13, or maybe one chart example with series + data, and one with dataTable + dataMapping?

File: docs/connectors/creating-your-first-chart.md (hunk 22 above)
Comment (kamil-musialowski) [reply docs/connectors/creating-your-first-chart.md:135]: Discussion in https://github.com/highcharts/connectors-morningstar/pull/194#discussion_r3879170390

File: docs/connectors/creating-your-first-chart.md (hunk 17 above)
Comment (kamil-musialowski) [reply docs/connectors/creating-your-first-chart.md:180]: Yes, it could be a good idea. If we're good with a screenshot, then not problem. If we want a live data, we need to either:
- change the config to match something that is available on demo live data server
- upload new dataset to server

What do you think?

File: docs/connectors/creating-your-first-chart.md (hunk 21 above)
Comment (pawelfus) [reply docs/connectors/creating-your-first-chart.md]: Correct, requires v13. However, `dataTable` option should be our preferred way for connectors. `series.data` was just intermediate solution, waiting for full DataTable support. Connectors produce DataTable, so using `dataTable` option is a natural choice. 

> So I'd suggest either a hint, that dataTable approach is available from v13, or maybe one chart example with series + data, and one with dataTable + dataMapping?

Other way around. We can add a hint for pre v13 versions that `series.data` should be used. Assuming the article will remain clear.

File: docs/connectors/creating-your-first-chart.md (hunk 17 above)
Comment (pawelfus) [reply docs/connectors/creating-your-first-chart.md:180]: I prefer real chart. We can switch to something that is already available.

**Review by pawelfus — APPROVED** [review]
Thanks!

File: docs/connectors/creating-your-first-chart.md [hunk 26]
```diff
@@ -0,0 +1,321 @@
+# Creating Your First Chart from Morningstar Funds Data
+
+Morningstar is one of the most trusted names in financial data. But getting that data onto a screen has traditionally meant handling complex requests, and a mapping layer that someone has to maintain forever.
+
+The Highcharts connectors for Morningstar Direct Web Services (DWS) remove that detour. You configure your credentials once, pick the data you want, and the connector hands you a formatted data table that Highcharts can render directly, with easy access to raw data as well.
+
+In this tutorial we will go from an empty HTML file to a working chart built on Morningstar data. We will use the DWS bundle, which targets Morningstar's newer API, and we will finish by looking at how the same few lines of setup unlock the rest of the DWS data catalogue.
+
```
mekhatria: Nice article :)
Is there a way to add either a picture, a gif, or a demo so people can see the end result?

kamil-musialowski: Sure thing! We already have one demo added to this article over here: https://github.com/highcharts/connectors-morningstar/pull/194/changes#diff-0264f84ceb125191f3fb1cef4a67a1ad7dce1f5542654972ba423c61918181eaR170

Would you like to add something more?

